PR #6, opened 2017-04-19
New UI

nabeelomer:
open


evilsocket:
I'm sorry but I had to revert this PR, I don't know if you really tested, but since the dialog is invoked from different threads, it doesn't really work well, all dialogs are stuck and lots of Gtk warnings on the console.

Issue #38, opened 2017-05-04
Do not use root (even for daemon)


adisbladis:
We can use the amazing Linux feature capabilities!

The only tricky bit is that the iptables CLI command is being called so ambient capabilities would have to be used.
Currently opensnitch would have to use:
1. cap_dac_override (to enable ftrace via sysfs)
2. cap_net_admin (nfqueue, iptables)

Obviously this is blocked by #20 since without this you would have to set the capabilites for the Python interpreter.

evilsocket:
Mmmm interesting, I know about linux caps but never used them honestly (i know i know, shame on me :D) ... so ideally we would just give such caps to the process and run it as a separate user?

adisbladis:
Yes exactly like that. Running with the bare minimum of privileges to do what we need.

eliasp:
IMHO `iptables` shouldn't be used, but `opensnitchd` should communicate with the Kernel through netlink instead. `iptables` is just an interface to the Kernel's packet filtering facilities, just in the same way `opensnitchd` will be.

When using `iptables`:
- every action has to be translated into an `iptables` command
- the execution of `iptables` involves a lot of context-switching
- complicates privilege separation/isolation
- execution of commands needs to be hardened against command injection
- using `iptables` is a one-way street

When using the Kernel's netlink interface:
- far more fine-grained control over actions
- make use of the nftables framework instead of the outdated netfilter framework
- place rules/hooks in the nftables Lua VM which act as a callback to `opensnitchd` to notify about events or trigger actions which require user-interactivity

The [netfilter project](https://www.netfilter.org/) also provides libraries to interact with the Kernel.

evilsocket:
Any doc. on how to interact from python with netlink?

eliasp:
There are a few Python netlink libraries, haven't worked with any of them, so I can't really tell how well they'd cover the required functionality, how well they're maintained, etc.

- https://github.com/facebook/gnlpy
- https://github.com/Robpol86/libnl
- http://docs.pyroute2.org/

strazzere:
Agreed with @eliasp that having a userland daemon communicate with a kernel would be best. This is actually how LittleSnitch operates on osx, userland daemon connects to a kext that catching all outgoing sockets. This allows you to catch /everything/ from your user and others, while alerting and allowed the current user to decide an action to take.

[...]

PR #41, opened 2017-05-05
Use iptables trough iptc, enables rootless operation

adisbladis:
Related to #38 

Provided I do `setcap 'cap_net_raw,cap_dac_override,cap_net_admin=+ep' $(which python2)` I can run opensnitch with an unprivileged user.

Needed `cap_net_raw` because of how communication with the kernel happens from iptables.

I say we still keep #38 open though, this is a larger issue

### DIFF
--- a/opensnitch/snitch.py
+++ b/opensnitch/snitch.py
@@ -16,12 +16,11 @@
 # program. If not, go to http://www.gnu.org/licenses/gpl.html
 # or write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
-import os
 import logging
 from netfilterqueue import NetfilterQueue
-from socket import AF_INET, AF_INET6, inet_ntoa
 from threading import Lock
 from scapy.all import *
+import iptc
 
 from opensnitch.ui import QtApp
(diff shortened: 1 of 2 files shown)


evilsocket:
is it dependant on this https://github.com/evilsocket/opensnitch/pull/43 ?

adisbladis:
No it's not

evilsocket:
Question: how do we execute setcap on first run in order not to have the user do it? Maybe we should require root for first run, setcap on ourself and then re-run without root?

adisbladis:
We don't. We would have to do setcap on the whole python interpreter which is bad for obvious reasons.
I see this more as a nice PoC (and much better than invoking external commands).

This should also translate fairly well to `libiptc` if we decide to go down that route.

evilsocket:
Not sure if I want to merge it on master tbh ...

adisbladis:
I think it's fine to leave this unmerged. It's not a very significant improvement.
This was mostly me wanting to explore what was needed to run something like this without running as root.

evilsocket:
:+1:

adisbladis:
I'm closing this, lets keep it in mind as a future reference :)

evilsocket:
absolutely, great job btw :+1:

PR #43, opened 2017-05-05
Python3 port

adisbladis:


evilsocket:
why? :)

adisbladis:
As said in the devchat
> PyQt5 only provides python3 wheels, so installation must be done from source
Being able to `pip install PyQt5` provides a much nicer development experience

I also happen to think it's a much nicer language and at some point Python2 is going to stop getting support

evilsocket:
mmm sounds cool then :)

PR #45, opened 2017-05-08
Use inotify to monitor changes in desktop files

adisbladis:
Also made the following smaller changes:
Don't use homerolled ini-parser
Follow XDG spec for icon locations
Minor PEP8 fixes (app.py is now fully PEP8)


evilsocket:
man I'm learning so much from your PRs :D thanks <3

PR #57, opened 2017-05-09
Added CLI argument for the database path.

jacobculley:
Not sure if people will want this but I quickly made it so that you can change the location of the database with an argument.

### DIFF
--- a/bin/opensnitch
+++ b/bin/opensnitch
@@ -22,11 +22,20 @@ import sys
 import logging
 import warnings
 
+from os.path import expanduser
+
 from optparse import OptionParser
+
+if 'SUDO_USER' in os.environ:
+    home = expanduser("~%s" % os.environ['SUDO_USER'])
+else:
+    home = expanduser("~%s" % os.environ['USER'])
+filename = os.path.join(home, "opensnitch.db")
 
 parser = OptionParser()
 parser.add_option( "--log-file",  dest="logfile", default=None, help="Log to file.", metavar="FILE" )
 parser.add_option( "--debug",     dest="debug", action="store_true", default=False, help="Enable debug logs." )
+parser.add_option( "--database",  dest="database", default=filename, help="Database path.", metavar="FILE" )
 
 (options, args) = parser.parse_args()
 
@@ -59,8 +68,9 @@ warnings.filterwarnings("ignore", category=RuntimeWarning, module="gtk")
 
 from opensnitch.snitch import Snitch
 
+
 def main():
-    snitch = Snitch()
+    snitch = Snitch(options.database)
 
     try:
         logging.info( "OpenSnitch v%s running with pid %d." % ( VERSION, os.getpid() ) )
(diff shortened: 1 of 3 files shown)


zero77:
Is this still an argument as it does not appear under help and fails to start when used.

themighty1:
@zero77 , no, this has been changed a long time ago. Nowadays the DB is fully in-memory for fast access. Do you have a use case to have the DB not be in-memory?

zero77:
@themighty1 
opensnitch uses a lot of memory over time and i wanted to see if using a db file would make a difference.

gustavo-iniguez-goya:
how much memory are we talking about? maybe it's a bug

I added the option to configure the DB from the GUI, but it's disabled right now. The amount of data written is huge.

themighty1:
yes, it would make a difference. All that data would be on the disk instead. For now, you can clean the db by pressing the small "delete all intercepted events" small button in the top right corner of the Events tab. This will free up the RAM used by the db.

zero77:
> how much memory are we talking about? maybe it's a bug

I will keep an eye on it and get back to you.

> I added the option to configure the DB from the GUI, but it's disabled right now. The amount of data written is huge.

Yes i saw that option, would you consider enabling it.

themighty1:
@gustavo-iniguez-goya , I remember I had some run away processes which would bombard the internet with endless connection attempts overnight. The db would run up into 100s of MB.

zero77:
> yes, it would make a difference. All that data would be on the disk instead. For now, you can clean the db by pressing the small "delete all intercepted events" small button in the top right corner of the Events tab. This will free up the RAM used by the db.

That didn't seem to make much if any of a difference but, i only cleared events so i could try clearing more.

gustavo-iniguez-goya:
ok, I've got the GUI writing events to disk. However I'm unable to swich between DBs without causing the GUI to crash. 
I can create a branch and maybe you @themighty1 can take a look ? otherwise the user should restart the GUI in order the change to take effect.

themighty1:
@gustavo-iniguez-goya , sure, I could take a look.

zero77:
@gustavo-iniguez-goya @themighty1 
What about having a threshold, where everything over that gets written to disk, or would this just cause instability.

gustavo-iniguez-goya:
@themighty1 , here it is: https://github.com/evilsocket/opensnitch/tree/ui-db-disk

@zero77 , it's not a bad idea. every n minutes backup data to disk. But for now let's start writing events to disk.

Once we do this, we'll have to take into account to do schema migrations if we update the DB tables.

themighty1:
@gustavo-iniguez-goya ,looks ok. The change of the db type will happen when the gui is restarted. I think that's the way it should be, no need to try to change it on the fly.
There should be a notification though that the change will take place the next time the gui starts.

zero77:
@themighty1 
What about giving users an option to restart from ui and informing them a restart is needed for this change to take effect.

themighty1:
@zero77 , I personally wouldnt spend my time on that feature. 
@gustavo-iniguez-goya , do you think it is something we would accept a PR for ?

gustavo-iniguez-goya:
I'd only add a notification to inform the user that the GUI must be restarted in order the change to take effect.

gustavo-iniguez-goya:
Changes added in 8a204c0

Adding options for MySQL, PostgreSQL would be very easy, we can also add it.

PR #60, opened 2017-05-10
feature/hidpi support

melizeche:
New UI for testing, works fine with hidpi resolutions, in lower resolutions the dialog is bigger than the original but all the info is displayed correctly and IMHO looks better.

### DIFF
--- a/opensnitch/resources/dialog_hi.ui
+++ b/opensnitch/resources/dialog_hi.ui
@@ -1,0 +1,229 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>Dialog</class>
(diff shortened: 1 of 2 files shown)


arielf:
@melizeche @evilsocket  Thanks so much for this fix!
(I was going to report it, but you all did a great job reporting and fixing ahead of me)
Looks great now even with the crazy-long chrome/chromium command lines.

in1t3r:
@arielf if you are using latest version of the master branch then you are actually using my changes on the dialog.ui :) I'm glad you like it. Its not the best I hope to find a way to resize it as much as it need and to not over resize it. But lets leave that for tomorrow. :)

PR #61, opened 2017-05-11
Fix for the resizing window and text issue

in1t3r:
I hope after this one there will be no more issues with text not resizing it works from small sized window to big size window.


arielf:
Thanks for all this work on the UI.

With latest, I still experienced one case of a very long (java with ~50 *.jar args) command that didn't fully fit in the window.  The good news is that this time, I could easily resize the window and see all the text, so I don't consider this a big problem anymore.

Issue #74, opened 2017-05-18
cannot do ping command while opensnitch is active


in1t3r:
cannot do ping command while opensnitch is active, your version in @evilsocket repo. In mine it works fine but my lack improvement of non blocking connections.
ways to reproduce start opensnitch run it open some apps allow some rules and then try pinging google.com for example or even IP not a fqdn.

adisbladis:
This is an issue with us not handling ICMP at all.
ICMP breaks some assumptions currently done in opensnitch (like not having sport and dport).

Here can be a good starting point for anyone interested in fixing this: https://github.com/evilsocket/opensnitch/blob/master/opensnitch/connection.py#L38

boozedog:
I'm seeing this problem too, running 1.4.0~rc4-1 amd64 . I cannot ping any host (including localhost) when OpenSnitch is active ... and nothing appears in the event tab to indicate that something is being blocked.

gustavo-iniguez-goya:
Please, install latest v1.4.0: https://github.com/evilsocket/opensnitch/releases/tag/v1.4.0

boozedog:
> Please, install latest v1.4.0: https://github.com/evilsocket/opensnitch/releases/tag/v1.4.0

That worked! Thank you.

alexfriman:
Same problem on v1.5.8
Linux 6.1.0-25-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.106-3 (2024-08-26) x86_64 GNU/Linux

PR #75, opened 2017-05-19
Add ICMP support

adisbladis:
Not very polished and can't find associated process (psutil does not have icmp support at all) but at least it' better than the current state where no ICMP comes through.


adisbladis:
This one fixes https://github.com/evilsocket/opensnitch/issues/74

Issue #20, opened 2017-05-01
Main logic and UI should be separated into a daemon and a dbus based UI process.


tidux:
Running an X application as root destroys any shred of security the Linux desktop might otherwise have.  A root-privileged daemon and an unprivileged user interface talking across a Unix socket or some other form of IPC would be far more reasonable for a security application.

voltagex:
Isn't this pretty much what Polkit is for?

See also https://news.ycombinator.com/item?id=14252265

hobarrera:
@voltagex Polkit is also for running GUI apps as root, this is what being advocated against here. Privilege separation is probably the cleanest approach.

evilsocket:
I'll repost this here, just to have more feedback :)

## Future Plan

* `opensnitchd` will be a C++ daemon, running as root with the main logic. It'll fix [this](https://github.com/evilsocket/opensnitch/issues/28).
* `opensnitch-ui` python (?) UI running as normal user, getting the daemon messages. Will fix [this](https://github.com/evilsocket/opensnitch/issues/20)
* `opensnitch-ruleman` python (?) UI for rule editing.

## Questions

1. What is the best IPC method in this case? I mean, if the daemon just creates a unix socket readable and writable by any user, any third party malicious software could access it and simply `ACCEPT` every packet ... dbus? No idea how it works honestly.

2. What's the best way in your opinion to keep all the involved/interested developers in sync and let them communicate without using github issues? Mailing list? Slack? Pigeons?

hobarrera:
In response to your questions:

1. Probably making the socket writable by a group, and then users that are allowed to control the daemon should be members of that group. That doesn't exclude malicious apps though, but I can't think of anything. This'll probably require some research.

2. GitHub issues tends to work the best for projects of this size and age (a mailing list is a bit unfriendly to sum, and you might keep people off by it, TBH). Pidgeons won't scale if you have remote developers. 😛

evilsocket:
@hobarrera Mmmm, about 1, from the bottom of my deep ignorance about UI and IPC on GNU/Linux, I was thinking about something like:

- Daemon runs a websocket (or similar streaming protocol) server with authentication on a unix socket.
- Authentication token/credentials are globally stored inside `/etc/opensnitch/auth.conf` or whatever.
- On first run, UI will ask the user for such credentials and use that to authenticate towards the daemon.

hobarrera:
[...]

Issue #65, opened 2017-05-12
Qt blocks Ctrl-C


adisbladis:
The Qt mainloop blocks and intercepts Ctrl-C.
This means Ctrl-C does not lead to the application being terminated.

This was always the case. This issue has just been amplified now that the Qt mainloop is always running.

allo-:
Same here. SIGTERM doesn't seem to reset the firewall rules. Is there currently any way to stop it without flushing iptables?

adisbladis:
@allo- This issue will largely be resolved when https://github.com/evilsocket/opensnitch/issues/20 is fixed since the rules should not be directly handled by the ui.

I tried looking into it but me and gui programming don't really mix.

adisbladis:
Fixed in https://github.com/evilsocket/opensnitch/pull/84

PR #84, opened 2017-06-12
Dbus ipc

adisbladis:
Please test properly before merging, this is a big one :)

Fixes https://github.com/evilsocket/opensnitch/issues/20 and https://github.com/evilsocket/opensnitch/issues/65


evilsocket:
can't test right now, I'll just merge since master is not supposed to be stable anyway :)

PR #91, opened 2017-06-28
Add Fedora build dependencies

da2x:
* Move dependencies from `Install` to `Requirements`
* List dependencies and sort alphabetically


adisbladis:
Forgot to say thanks yesterday so... Thanks!

PR #115, opened 2018-04-06
Added go get insutrctions to obtain source and go libraries required …

medoix:
This could be pretty basic for most but as someone who has never used GO i needed to figure this out as it was throwing errors when trying to build after simply checking out the git repo.

Feel free to decline or ask for changes.


evilsocket:
The daemon is using glide, it'd be better to have those instructions instead.

PR #133, opened 2018-04-10
added required libraries in README

hohenstaufen:
I have added in the README some required libraries to get opensnitchd working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,13 @@
 
 ### TL;DR
 
-    sudo apt-get install protobuf-compiler libpcap-dev libpcapnetfilter-queue-dev python-pyqt5
+    sudo apt-get install golang protobuf-compiler libpcap-dev libnetfilter-queue-dev python-pyqt5
     go get github.com/golang/protobuf/protoc-gen-go
+    go get github.com/fsnotify/fsnotify
+    go get github.com/google/gopacket
+    go get github.com/google/gopacket/layers
+    go get google.golang.org/grpc
+    go get golang.org/x/net/context
     go get -v github.com/Masterminds/glide
     cd /path/to/this/repo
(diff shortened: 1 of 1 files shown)


hohenstaufen:
Also, about the `python-pyqt5` package, it seems it is too much effort for the devs to put a python2 version of PyQt5 on pypi (see https://riverbankcomputing.com/pipermail/pyqt/2017-October/039676.html).  
So, the alternative would be switching to python3; I tried it by just updating package requirements, but the code is not python3-ready.  
That is what the first error looks like:
```Traceback (most recent call last):
  File "./opensnitch-ui", line 36, in <module>
    service = UIService(app, on_exit)
  File "/home/alberto/opensnitch/ui/service.py", line 41, in __init__
    self._setup_interfaces()
  File "/home/alberto/opensnitch/ui/service.py", line 55, in _setup_interfaces
    names = array.array('B', '\0' * bytes)
TypeError: cannot use a str to initialize an array with typecode 'B'```
I haven't put any more effort into it since python2 may be an explicit design choice.

File: README.md [hunk 1]
```diff
@@ -16,8 +16,13 @@
 
 ### TL;DR
 
-    sudo apt-get install protobuf-compiler libpcap-dev libpcapnetfilter-queue-dev python-pyqt5
+    sudo apt-get install golang protobuf-compiler libpcap-dev libnetfilter-queue-dev python-pyqt5
     go get github.com/golang/protobuf/protoc-gen-go
```
neunkasulle: Suggestion: Instead of listing the packets here, a `go get .` in the daemon folder should also do the trick

hohenstaufen: It was my first time with golang, so i did not know there was a better solution 😅

PR #143, opened 2018-04-14
Update README.md for redhat based systems

intika:
Update install instruction for redhat based systems


intika:
Amazing work by the way !

evilsocket:
thanks but this is not redhat specific stuff

PR #148, opened 2018-04-14
Disable timer on mouse click

intika:
Disable timer on mouse click 
- on frame click 
- on combo-box select click


evilsocket:
Do **NOT** send PR about changes of the core logic, they will be rejected.

PR #155, opened 2018-04-15
I created a tl;dr for installing opensnitch on Fedora

sirredbeard:
Here is what I learned:
Tried to shift protoc-gen-go and godep to versions directly from fedora repos.
Discovered protoc-gen-go and protobug-compiler packages clashed, filed bug https://bugzilla.redhat.com/show_bug.cgi?id=1567650
Also discovered python3-pip package needed to be updated in repos, filed bug https://bugzilla.redhat.com/show_bug.cgi?id=1567652
Figured out had had to put project in $HOME/go/src, see golang/dep#148
No need to install anything via pip ahead of time, now the specific packages needed are all installed via requirements.txt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,24 @@
   <img src="https://raw.githubusercontent.com/evilsocket/opensnitch/master/screenshot.png" alt="OpenSnitch"/>
 </p>
 
-### TL;DR
+### TL;DR - Debian/Ubuntu
 
     sudo apt-get install golang protobuf-compiler libpcap-dev libnetfilter-queue-dev
+    python3 -m pip install --user grpcio-tools
+    go get github.com/golang/protobuf/protoc-gen-go
+    go get -u github.com/golang/dep/cmd/dep
+    cd /path/to/this/repo
+    make
+    sudo make install
+
+    sudo systemctl enable opensnitchd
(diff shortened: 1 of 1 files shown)


sirredbeard:
Closing until figured out some issues with UI on Fedora.

evilsocket:
What issues if I may ask?

Herve5:
Evilsocket, don't be too bad against sirredbeard : you are now so successful that many new users may be like me : very light on computing, with difficulties even in compiling OpenSnitch...
I for one remember using it successfully last year (on Ubuntu Mate) but having difficulties in duplicating what I did before on OSX with LittleSnitch; I just tried to upgrade to the present version... and didn't manage to get the supporting libraries. This gives you the extent of my incompetence :-)
But, you will have more and more naive users from now on :-)

sirredbeard:
Everything installed.
The rule-creation prompts appeared.
But I couldn't get the rule-editing GUI to launch.
Also every time the rule-creation prompt appeared it also caused a big, oversized "opensnitch is ready" notification appears in gnome. It can be fixed by turning off notifications for opensnitch or by installing this [extension](https://extensions.gnome.org/extension/1007/window-is-ready-notification-remover/).

PR #169, opened 2018-04-28
UI Updates & Function to Disable Auto Apply

medoix:
I probably should have split these changes the UI updates may have something to do with me only having HIDPI displays (4k monitors + laptop). I also updated so the window can be resized and the content stretches accordingly.

The second change was to have the ability to disable the "Auto Apply" functionality as is the default in Little Snitch (auto allowing the rules kind of defeats the purpose).

If the UI changes cause problems on non HIDPI displays then i am happy to split this PR.


evilsocket:
there's a reason why things have been made a certain way (for instance: WindowStaysOnTopHint), you should have asked before just modifying stuff

medoix:
You ask for help on twitter constantly and i have tried a couple of times to reach out to you (not sure where to talk to you but have tried keybase) but to be honest you come across super arrogant and not welcoming at all (hopefully just a cultural difference) but if not all the best and ill just keep my own fork.

medoix:
P.S i was not offended just providing feedback, i have never attempted to chat on twitter (just follow you) and find it curious that you said communicate here on github which is what i'm doing...

evilsocket:
>  i can clearly see you are talented.

it's not about being talented but about knowing better how to organize a project

> you provide no feedback and close the requests instantly

I told you why I was closing the PR in the very first reply, you don't like my feedback != I don't provide feedback.

>  close the requests instantly

do not generalize, it's just yours.

> No were on the project does it state to open issues for questions

because that's common sense and shouldn't be stated anywhere ... the several thousands of other users seem to have clearly understood that without any "official statement".

> issues were actuality locked last i checked to people who had contributed so i couldn't

issues are only locked when they're offtopic, what you're saying is incorrect at best.

> It is highly likely my lack of experience working on open source projects

yes

> i was merely seeking feedback as i said in the PR message.

no, you did not, you only said "i can split the PR", which is different than asking feedback, and anyways, feedback should be asked before the PR, not after, otherwise there's a chance it might not get merged.

> i was not offended just providing feedback, i have never attempted to chat on twitter (just follow you) 

than please, keep your assumptions and comments about my arrogance and persona for yourself given you apparently lack the correct amount of facts to judge :)

> you said communicate here on github which is what i'm doing...

all you're doing **now**, after the PR, I already explained that shouldn't be the case.

Issue #184, opened 2018-07-27
Kprobe error


StevenDoesStuffs:
Log: 
```
[root@StevenPC opensnitch-git]# opensnitchd
[2018-07-27 16:35:32]  IMP  Starting opensnitch-daemon v1.0.0b
[2018-07-27 16:35:32]  !!!  Error while enabling probe descriptor for opensnitch_exec_probe: write /sys/kernel/debug/tracing/kprobe_events: no such file or directory
```
I've checked that in fact the file does exist.

Northern-Lights:
I'm seeing the same on Fedora 28.

OderDochNicht:
Confirmed for 4.17.0-1-amd64 #1 SMP Debian 4.17.8-1 (2018-07-20) x86_64 GNU/Linux

DickThomas:
same for 4.17.11-arch1 #1 SMP PREEMPT Sun Jul 29 10:11:16 UTC 2018 x86_64 GNU/Linux.
not sure what else I can add.

OderDochNicht:
Not here, unfortunately:

`sudo /usr/local/bin/opensnitchd 
[2018-08-15 11:25:54]  IMP  Starting opensnitch-daemon v1.0.0b
[2018-08-15 11:25:54]  INF  Loading rules from /home/user/go/src/github.com/evilsocket/opensnitch/daemon/rules ...
[2018-08-15 11:25:54]  !!!  Path '/home/user/go/src/github.com/evilsocket/opensnitch/daemon/rules' does not exist.
user@monster:~/go/src/github.com/evilsocket/opensnitch/daemon$ mkdir rules
user@monster:~/go/src/github.com/evilsocket/opensnitch/daemon$ sudo /usr/local/bin/opensnitchd 
[2018-08-15 11:26:34]  IMP  Starting opensnitch-daemon v1.0.0b
[2018-08-15 11:26:34]  INF  Loading rules from /home/user/go/src/github.com/evilsocket/opensnitch/daemon/rules ...
[2018-08-15 11:26:34]  !!!  Error while creating queue #0: Error binding to queue: operation not permitted

user@monster:~/go/src/github.com/evilsocket/opensnitch/daemon$ ln -s rule/* rules
user@monster:~/go/src/github.com/evilsocket/opensnitch/daemon$ sudo /usr/local/bin/opensnitchd 
[2018-08-15 11:33:33]  IMP  Starting opensnitch-daemon v1.0.0b
[...]

PR #188, opened 2018-08-12
Update probed function sys_execve to do_execve

Northern-Lights:
This is to address #184: https://github.com/evilsocket/opensnitch/issues/184

`sys_execve` does not appear to be an available filter function:

```
# grep execve /sys/kernel/debug/tracing/available_filter_functions 
audit_log_execve_info
do_execveat_common.isra.37
__ia32_compat_sys_execve
__ia32_compat_sys_execveat
__ia32_sys_execve
__ia32_sys_execveat
__x64_sys_execve
__x64_sys_execveat
do_execve
do_execveat
```
In order to fix this, I tried `__x64_sys_execve` and `do_execve`.  Both appear to allow the daemon to proceed with initialization and to properly operate the packet filter.  The UI connects to the daemon and pops up windows upon alerts with no issue.

### DIFF
--- a/daemon/conman/connection.go
+++ b/daemon/conman/connection.go
@@ -71,7 +71,7 @@ func NewConnection(nfp *netfilter.Packet, ip *layers.IPv4) (c *Connection, err e
 	c = &Connection{
 		SrcIP:   ip.SrcIP,
 		DstIP:   ip.DstIP,
-		DstHost: dns.HostOr(ip.DstIP, ""),
+		DstHost: dns.HostOr(ip.DstIP, ip.DstIP.String()),
 		pkt:     nfp,
 	}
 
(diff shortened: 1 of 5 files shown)


Northern-Lights:
If changing the syscall will break compatibility, I could make a slice of possible syscalls, e.g. `[]string{"sys_execve", "do_execve"}`.  Then, if an error is returned on the first one, we can move on and try the next.

evilsocket:
need some time to test this, did you go for the slice approach?

Northern-Lights:
Sorry, I saw your comment a little late.  It is not currently trying both in a slice, although I see you've already merged the PR.  I'll be happy to add it in if anything pops up in the issues section.

Issue #197, opened 2018-09-07
opensnitch-ui crashes (aborts) when on_update_triggered is called


MartynHare:
How-I-managed-to-make-this-software-work-patch: [fix-abort-on-getpwall.patch.txt](https://github.com/evilsocket/opensnitch/files/2362442/fix-abort-on-getpwall.patch.txt)

Basically after building opensnitch, I found the UI would crash after every single prompt.  I managed to get the software to work by making the attached (above) changes to stats.py

--- First problem ---
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/opensnitch/dialogs/stats.py", line 209, in _on_update_triggered
    by_users["%s (%s)" % (pw_name, uid)] = hits
UnboundLocalError: local variable 'pw_name' referenced before assignment

--- Second problem ---
Initialising pw_name earlier on in the code with the default error would get round the first problem but then I'd encounter this instead....

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/opensnitch/dialogs/stats.py", line 206, in _on_update_triggered
    pw_name = pwd.getpwall().pw_name
AttributeError: 'list' object has no attribute 'pw_name'
Aborted (core dumped)

MartynHare:
Oh.. I almost forgot to say thank you by the way.  I've been looking for a solution to do per-application outbound filtering on Linux for a while.  Good old Douane would kernel panic and although there are some other solutions out there (Subgraph OS firewall), this one seems to be the most simple to deploy.

Once I've gotten my head round some of the code (hint: not been a developer since Uni), I will see if this can be adapted to be packaged for Fedora as an additional firewall option! =D

luzpaz:
@MartynHare Perhaps you should make a PR from the patch above and mention
`Closes #197` in the git summary. It's easier to inspect the code instead of downloading the patch and opening it in a text editor etc...

MiWCryptAnalytics:
Also ran into this one, I closed the duplicate PR #202 
Could we please merge this, the ui is broken without this fix

PR #198, opened 2018-09-08
Fixes issue #197 - modified user stats to use pwd.getpwuid 

MartynHare:
This patch is to fix a problem where the UI would crash out whenever a dialog was responded to, first complaining that pw_name wasn't initialised, then that pwd.getpwall() doesn't normally take arguments. See #197 for more information.


MartynHare:
That suggestion works for me, tested it just now with everything working as expected :-)

PR #202, opened 2018-09-18
change pwall to pwuid (runtime error)

MiWCryptAnalytics:
The getpwall takes no arguments, and will cause a runtime error if called this way. 
I was getting a runtime unbound variable pw_name that was crashing the app.
The intent is to get user account for a given uid, so change to use pwuid function.


MiWCryptAnalytics:
duplicate #198

PR #229, opened 2018-12-13
Fixes #228: Parse IP and port to uint

### LINKED ISSUE #228: Daemon killed/restarts periodically
It happens every few seconds/minutes on Ubuntu 16.04

screenshot from 2018-12-10 15-10-56

```Error while parsing address to int: strconv.ParseInt: parsing "other address": value out of range```

DarkSpyro003:
Fixes a crash caused by trying to parse too large hexadecimal numbers into signed ints. Parses to uint instead.
For issue #228

### DIFF
--- a/daemon/conman/connection.go
+++ b/daemon/conman/connection.go
@@ -18,9 +18,9 @@ import (
 type Connection struct {
 	Protocol string
 	SrcIP    net.IP
-	SrcPort  int
+	SrcPort  uint
 	DstIP    net.IP
-	DstPort  int
+	DstPort  uint
 	DstHost  string
 	Entry    *netstat.Entry
 	Process  *procmon.Process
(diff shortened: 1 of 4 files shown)


datio:
Thanks for trying to tackle this! I installed opensnitch with your changes but it still throws a fatal error:

screenshot from 2018-12-16 13-54-44

cwh1te:
@datio that seems strange - afaict all the calls to `strconv.ParseInt` were changed to `strconv.ParseUint` by this patch so I'm not sure where those errors could be coming from in your screenshot.

datio:
Yes indeed, thank you for noticing! it seems I did something wrong when applying the patch.
I reinstalled it using @DarkSpyro003's repo and it worked like a charm.

Issue #205, opened 2018-09-21
opensnitch-gui  high CPU usage


lyhixuso:
Latest opensnitch-ui pull unusable in Ubuntu Mate 18.04.

Max CPU usage as soon as scrolling is attempted on the general tab.  Clicking other  tabs also max CPU.   stdio repeats 

> 
> QApplication: invalid style override passed, ignoring it.
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.6/dist-packages/opensnitch/dialogs/stats.py", line 205, in _on_update_triggered
>     pw_name = pwd.getpwall(int(uid)).pw_name
> TypeError: getpwall() takes no arguments (1 given)
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.6/dist-packages/opensnitch/dialogs/stats.py", line 209, in _on_update_triggered
>     by_users["%s (%s)" % (pw_name, uid)] = hits
> UnboundLocalError: local variable 'pw_name' referenced before assignment
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.6/dist-packages/opensnitch/dialogs/stats.py", line 205, in _on_update_triggered
>     pw_name = pwd.getpwall(int(uid)).pw_name
> TypeError: getpwall() takes no arguments (1 given)
> 
> 
> 
>

savoury1:
Patching source with pull request #198 was first necessary for me to even load opensnitch-ui on my system or same exceptions mentioned above occurred immediately on attempting to start the UI, which for me failed to load at all without applying #198 patch.

After patching source and building package again opensnitch-ui loads successfully (notification icon is visible) and prompts do start appearing to allow/deny traffic. Similar to above comment, when opening the UI to view statistics one CPU core immediately hits 100% usage (process opensnitch-ui) whenever the General tab is selected (which is the default tab on first opening the UI).

However, in my case (with #198 applied) clicking any other tabs results in CPU usage for opensnitch-ui returning to usual levels (4-5% CPU), so this is distinct from the comment above. Thus it appears there is an issue specifically with the General tab of opensnitch-ui that is causing the high CPU usage.

lyhixuso:
I can replicate all your issues. The General Tab maxes the CPU.   CPU consumption drops after clicking on other tabs.  All other tabs are empty except General Tab.

gustavo-iniguez-goya:
Hi @lyhixuso @savoury1 ,

[...]

PR #264, opened 2019-06-16
Use QTableView instead of QTableWidget to render statistic data

gustavo-iniguez-goya:
This change drops down the CPU usage on my PC from 100% to 2-5%. It would be a good idea to test it in other evironments #205


Ph0rkoz:
I get a bunch of errors in ui.pb.go when I merge this request. And it fails to build the daemon.

Ph0rkoz:
figured out it pulls real source from the  main repo when compiling.. and this didn't help on my machine. CPU usage is still high when the UI is opened.

Issue #262, opened 2019-06-10
[Question] How to block all connections by default?


lyndalopez544:
I want opensnitch to block all connections with any pattern by default and then create whitelisting rules.

d-bykov:
Yeah, I also stuck with this problem. I can set the `"default_action"` parameter to `"deny"` in `~/.opensnitch/ui-config.json`, but sadly it only works when GUI is running. If `opensnitch-ui` isn't running, all connection that are not defined in rules files are allowed by default.

It would be great if there would be an option to comletely block any connections not defined in rules by default when GUI isn't running.

gustavo-iniguez-goya:
The problem is that the default behaviour is hardcoded here: https://github.com/evilsocket/opensnitch/blob/master/daemon/ui/client.go#L23

I've modified it to load a config file from /etc/opensnitchd/deault-config.json with a configuration to deny all connections once. This way when as soon as the GUI is running again it'll prompt the user to perform an action.

Would be this solution good enough? We'll still have to modify it by hand and we'll have 2 config files, but I think it's ok to have it separately.

d-bykov:
> The problem is that the default behaviour is hardcoded here: https://github.com/evilsocket/opensnitch/blob/master/daemon/ui/client.go#L23
> 
> I've modified it to load a config file from /etc/opensnitchd/deault-config.json with a configuration to deny all connections once. This way when as soon as the GUI is running again it'll prompt the user to perform an action.
> 
> Would be this solution good enough? We'll still have to modify it by hand and we'll have 2 config files, but I think it's ok to have it separately.

Looks promising to me.

gustavo-iniguez-goya:
PR proposed. Note that currently the rule/field duration is ignored, only the default action is applied due to how the Ask reply is handled. I don't know if it would be useful for someone or in some use case to save all denied connections to disk.

gustavo-iniguez-goya:
fixed with latest version.

fabianski7:
for this to work is it necessary for the GUI to be open?
the rules I create do not work when it is closed, even if the daemon is running

gustavo-iniguez-goya:
nope, it's not necessary. What version are you using, 1.3.6 or did you compile it from sources?

Please, set LogLevel to 0 in `/etc/opensnitchd/default-config.json` and paste the log, indicating what rule should apply in what case, etc.  Thank you!

fabianski7:
I think I understand why.
[...]

PR #269, opened 2019-07-02
Allow to configure daemon's default action rule

gustavo-iniguez-goya:
If the file /etc/opensnitchd/default-config.json exists,
read it and apply the options to the default rule when there's no client
connected.

If it doesn't exist, just apply the default rule, allow connections
once.

Config example: {"default_action": "deny", "default_duration": "once"}

thread #262


MoreUser:
If you change 3 lines at daemon/ui/client.go it will compile.
Line 62, 64 and 70: "confFile" to "configFile"

PR #277, opened 2019-08-13
Fixed unwanted rules reloading after adding a new one

### LINKED ISSUE #271: Temporary rules resetted after on disk rules are reloaded
If the rules stored in /etc/opensnitchd/rules/ change, the temporal (in memory) rules are deleted:

https://github.com/evilsocket/opensnitch/blob/master/daemon/rule/loader.go#L64

After a rule is added, changed or removed, as the temporary rules are deleted, the user will be prompted to allow/deny them again.

gustavo-iniguez-goya:
When adding a new permanent rule, the temporary ones are deleted, so the
user is prompted again to allow/deny connections.

fixes #271


Ph0rkoz:
broken :(

PR #290, opened 2020-02-02
some python3 fixes

collinanderson:


gustavo-iniguez-goya:
hey @collinanderson , thank you for sending this. 

Now we have the ability to configure ads domains from the GUI, so this script is deprectated.

PR #296, opened 2020-06-04
Catch 404 exceptions, update to python 3 syntax

m-j-r:


gustavo-iniguez-goya:
This script is deprectated. Now we have the ability to configure ads domains from the GUI.

Thank you for sending it anyway!

PR #301, opened 2020-11-13
typo: cast -> case

raboof:


gustavo-iniguez-goya:
@raboof thank you for your PR!

Things have changed a little bit lately, and the typo is not longer in the README. I moved it to the wiki here: https://github.com/gustavo-iniguez-goya/opensnitch/wiki/Rules

And thanks that you spotted this typo I fixed it there :)

PR #306, opened 2020-12-18
    Add a custom view/model to display the connections list more effi…

themighty1:
…ciently than currently with QSqlQueryModel.

    The major steps are:

    - take advantage of sqlite's default autoincremented rowid column
    - add index of the db columns
    - when a filter is applied, we build a map of rowids corresponding to the filter
    - when user scrolls the view, query the db only for that portion of db which contains the rows to be displayed
    - because sqlite cannot use an index when a wildcard is at the start of the LIKE expression, e.g. "process LIKE '%sbin%'", use a workaround:
      - keep track of all distinct values in each column
      - check in python if any of the distinct values contain the filter string
      - reconstruct the query string: instead of "process LIKE '%sbin%'" we use "process IN (<list of distinct values containing 'sbin'>)"

    Minor steps:
    - mimic some QSqlQueryModel's methods so that our model can be a drop-in replacement
    - disable view's default scrollbar and use our own scrollbar which is aware of how many rows are in the db

### DIFF
--- a/ui/opensnitch/customwidgets.py
+++ b/ui/opensnitch/customwidgets.py
@@ -1,6 +1,10 @@
 from PyQt5 import Qt, QtCore
-from PyQt5.QtGui import QColor, QPen, QBrush
-from PyQt5.QtSql import QSqlDatabase, QSqlQueryModel
+from PyQt5.QtGui import QColor, QPen, QBrush, QStandardItemModel, QStandardItem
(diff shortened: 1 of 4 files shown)


themighty1:
I've been running this code for a few days with over 1 million rows in the db.

gustavo-iniguez-goya:
It works really well @themighty1 , really really smooth, awesome!

PR #310, opened 2020-12-19
update systemd service file

deathtrip:
It's better to let directory creation be handled by packaging, instead of executing it on every daemon start. On Arch Linux for example, that directory is already contained in the package.
When that line is removed `PermissionsStartOnly=true` can go as well, it's [deprecated](https://github.com/systemd/systemd/pull/10802) anyway.

### DIFF
--- a/daemon/opensnitchd.service
+++ b/daemon/opensnitchd.service
@@ -6,8 +6,6 @@ After=network.target
 
 [Service]
 Type=simple
-PermissionsStartOnly=true
-ExecStartPre=/bin/mkdir -p /etc/opensnitchd/rules
 ExecStart=/usr/local/bin/opensnitchd -rules-path /etc/opensnitchd/rules
 Restart=always
 RestartSec=30


tianon:
FWIW, this should be doable in Debian packaging too most easily/directly via `debian/<package>.dirs`: https://manpages.debian.org/buster/debhelper/dh_installdirs.1.en.html

(I had a similar "huh, that's interesting" moment while reviewing this service file recently :smile:)

gustavo-iniguez-goya:
The PR can not be accepted as is, because we'd break non-arch installations.

I think that this task should be done from the daemon, otherwise we'll depend on the service manager used by the distro (systemd, sysv, s6, openrc, etc, etc)

deathtrip:
I don't think removing these lines would break opensnitch. As right now on non-systemd distros, the rules directory still has to be created some other way.

deathtrip:
As it's superseded by https://github.com/evilsocket/opensnitch/pull/1019, i'm  gonna close this one.

PR #313, opened 2020-12-22
Use daemon's event timestamp in UI.

themighty1:
- Use daemon's event timestamp in UI.
- Use the timestamp instead of the event object when iterating over the last events. ~15x speed increase.
- Increase event buffer to 100. On my machine I routinely hit the ceiling of 50 events under some multitasking workloads which results in new connections not being logged.

### DIFF
--- a/daemon/statistics/event.go
+++ b/daemon/statistics/event.go
@@ -27,5 +27,6 @@ func (e *Event) Serialize() *protocol.Event {
 		Time:       e.Time.Format("2006-01-02 15:04:05"),
 		Connection: e.Connection.Serialize(),
 		Rule:       e.Rule.Serialize(),
+		Unixnano:   e.Time.UnixNano(),
 	}
 }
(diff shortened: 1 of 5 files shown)


gustavo-iniguez-goya:
Thank you @themighty1 , please, commit also the file https://github.com/evilsocket/opensnitch/blob/master/daemon/ui/protocol/ui.pb.go

and https://github.com/evilsocket/opensnitch/blob/master/ui/opensnitch/ui_pb2.py https://github.com/evilsocket/opensnitch/blob/master/ui/opensnitch/ui_pb2_grpc.py

Otherwise the build check will fail.

themighty1:
Oh, so even those those files are generated by make, we still have to commit them every time ui.proto changes? 
Is there a reason why this is so?

gustavo-iniguez-goya:
yes, at least daemon/ui/protocol/ui.pb.go , in order the build checks work.

PR #326, opened 2021-01-04
german translation update

freddii:


gustavo-iniguez-goya:
I'll add new strings soon, so update the PR once I commit them. I'll post a comment here.

gustavo-iniguez-goya:
done @freddii , new strings added https://github.com/evilsocket/opensnitch/blob/master/ui/opensnitch/dialogs/prompt.py#L322

In the future I'll update all the translations, so you only have to translate the new ones.

freddii:
i made a new fork with the changes.

PR #328, opened 2021-01-05
Add Brazilian Portuguese translation

tioguda:


gustavo-iniguez-goya:
hi @tioguda !

Thank you for the translation. I updated some strings yesterday, so some translations are not translated properly, for example:
https://github.com/evilsocket/opensnitch/blob/e76c83cb1ab0f870248fefb1d374f7c8e5890783/ui/i18n/locales/pt_BR/opensnitch-pt_BR.ts#L1200

image

Could you pull latest changes from master, execute `make`, update the translations and check that all strings are translated properly?

Also maintain the original text(html) format of the strings, otherwise the text won't be displayed correctly in some situations.
For example:

`is attempting to resolve &lt;b&gt;%s&lt;/b&gt; via %s, %s port %d` should be:
`está tentando resolver  &lt;b&gt;%s&lt;/b&gt; via %s, %s porta %d` instead of `está tentando resolver via %s, %s porta %d`

**NOTE:** notice that the original string has 4 parameters (%s %s %s %d), and you only added 3. This causes GUI close unexpectedly. I'll change it to not depend of the variable type, but for now use the correct number of modifiers.

https://github.com/evilsocket/opensnitch/blob/e76c83cb1ab0f870248fefb1d374f7c8e5890783/ui/i18n/locales/pt_BR/opensnitch-pt_BR.ts#L1205
https://github.com/evilsocket/opensnitch/blob/e76c83cb1ab0f870248fefb1d374f7c8e5890783/ui/i18n/locales/pt_BR/opensnitch-pt_BR.ts#L1246
https://github.com/evilsocket/opensnitch/blob/e76c83cb1ab0f870248fefb1d374f7c8e5890783/ui/i18n/locales/pt_BR/opensnitch-pt_BR.ts#L1253
https://github.com/evilsocket/opensnitch/blob/e76c83cb1ab0f870248fefb1d374f7c8e5890783/ui/i18n/locales/pt_BR/opensnitch-pt_BR.ts#L1200
https://github.com/evilsocket/opensnitch/blob/e76c83cb1ab0f870248fefb1d374f7c8e5890783/ui/i18n/locales/pt_BR/opensnitch-pt_BR.ts#L1195

thank you!

tioguda:
I added all the text(html), I'm running the program and the strings are being translated.

If something is still not in compliance, let me know and I'll adjust it as necessary.

Greetings.

gustavo-iniguez-goya:
Now yes, thank you!

PR #342, opened 2021-01-31
maintain a cache of struct Process for currently active PIDs

themighty1:
decreases PID lookup time from ~100usec to ~5usec

### DIFF
--- a/daemon/procmon/activepids.go
+++ b/daemon/procmon/activepids.go
@@ -1,0 +1,86 @@
+package procmon
+
+import (
+	"fmt"
+	"github.com/evilsocket/opensnitch/daemon/log"
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+type value struct {
+	Process *Process
+	//Starttime uniquely identifies a process, it is the 22nd value in /proc/<PID>/stat
+	//if another process starts with the same PID, it's Starttime will be unique
+	Starttime uint32
+}
+
+var (
+	activePids     = make(map[uint32]value)
+	activePidsLock = sync.RWMutex{}
+)
+
+//monitorActivePids checks that each process in activePids
+//is still running and if not running (or another process with the same pid is running),
+//removes the pid from activePids
+func monitorActivePids() {
+	for {
+		time.Sleep(time.Second)
+		activePidsLock.Lock()
+		for k, v := range activePids {
+			data, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/stat", k))
+			if err != nil {
+				//file does not exists, pid has quit
+				delete(activePids, k)
+				continue
+			}
+			startTime, err := strconv.Atoi(strings.Split(string(data), " ")[21])
+			if err != nil {
+				log.Error("Could not find or convert Starttime. This should never happen. Please report this incident to the Opensnitch developers.")
+				delete(activePids, k)
+				continue
+			}
+			if uint32(startTime) != v.Starttime {
+				//extremely unlikely: the original process has quit and another process
+				//was started with the same PID - all this in less than 1 second
+				log.Error("Same PID but different Starttime. Please report this incident to the Opensnitch developers.")
+				delete(activePids, k)
+				continue
+			}
+		}
+		activePidsLock.Unlock()
+	}
+}
+
+func findProcessInActivePidsCache(pid uint32) *Process {
+	activePidsLock.Lock()
+	defer activePidsLock.Unlock()
+	if value, ok := activePids[pid]; ok {
+		return value.Process
+	}
+	return nil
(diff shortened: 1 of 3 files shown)


themighty1:
I've been running this patch for many days.

gustavo-iniguez-goya:
:+1: 

I think we need to move:
https://github.com/evilsocket/opensnitch/pull/342/files#diff-7546f8dbef78d251cddff3e77702a4b99118887b0b94c354f4e55211658fc256R110

after here:
https://github.com/evilsocket/opensnitch/blob/f881cee5438b97213bac7477c4ce2267ee6a413b/daemon/procmon/parse.go#L91

Otherwise cached PIDs using audit won't be cached. Apart from that, it works great.

themighty1:
I dont I think I understood your comment.
findProcessInActivePidsCache works only for method==proc
I dont see how it will affect the audit method?

themighty1:
Oops, sorry, I wasnt thinking straight. Ignore my last comment.

gustavo-iniguez-goya:
no problem :)

I think we can reorder FindProc() like these:
```
func FindProcess(pid int, interceptUnknown bool) *Process {
    if interceptUnknown && pid < 0 {
        return NewProcess(0, "")
    }

    if proc := findProcessInActivePidsCache(uint32(pid)); proc != nil {
        return proc
    }

    linkName := fmt.Sprint("/proc/", pid, "/exe")
    if _, err := os.Lstat(linkName); err != nil {
        return nil
    }

    if link, err := os.Readlink(linkName); err == nil {
        proc := NewProcess(pid, link)
        proc.readCwd()
        proc.readCmdline()

        if methodIsAudit() {
            if aevent := audit.GetEventByPid(pid); aevent != nil {
                audit.Lock.RLock()
                proc.Path = aevent.ProcPath
                proc.setCwd(aevent.ProcDir)
                audit.Lock.RUnlock()
                // if the proc dir contains non alhpa-numeric chars the field is empty
                if proc.CWD == "" {
                    proc.readCwd()
                }
            }
        }

        proc.readEnv()
        proc.cleanPath()

        addToActivePidsCache(uint32(pid), proc)
        return proc
    }
    return nil
}

```

and process.go, Init():

```
            go audit.Reader(auditConn, (chan<- audit.Event)(audit.EventChan))
+            go monitorActivePids()

```

gustavo-iniguez-goya:
oops, nevermind, I've realized why I created a new Process{} if the method was audit. [We can't rely on /proc for anything](https://sysdig.com/blog/hiding-linux-processes-for-fun-and-profit/), if there's a better alternative.

themighty1:
I was planning to revisit this PR today but I see it was merged.
So your initial suggestion was to let the audit method take advantage of this new active PIDs cache, right?
Honestly, I didn't get around to understanding how audit method works.

Does audit method rely on func newConnectionImpl()? We have to do the same /proc lookup for method "audit" as for method "proc" ? I'm sure I'm missing something. 

Does audit method keep a map of correlation between a PID and the IP-tuple from the connect() syscall? I could not find it.

gustavo-iniguez-goya:
Sorry, I'd like to release a new version with latest changes ASAP. Your fix for VPNs and the fixes for regexp rules are very important.

> So your initial suggestion was to let the audit method take advantage of this new active PIDs cache, right?

Yes, we're creating a NewProcess() object on FindProcess() that it's not being cached:
https://github.com/evilsocket/opensnitch/blob/master/daemon/procmon/parse.go#L95
So it needs also to be cached.

The reason for exiting from FindProcess() before checking the `/proc/<pid>/exe` link and read the cmdline file is that by using audit we already received the path of the process from the kernel directly, via the audit framework. That's why it's a little bit safer/reliable to use it, but still, we need to lookup the inode of the connection under /proc. There's no way to match an outgoing connection against the received audit events. And the audit framework does not log the inode of a connection, nor the source port of it (only srcIP -> dstIP:port).

> Does audit method rely on func newConnectionImpl()?

Yes, it works in the same way than ftrace: 
https://github.com/evilsocket/opensnitch/blob/master/daemon/procmon/parse.go#L62

Since it reads events (connect(), socket(), etc), from a pipe, it also has its own cache of pids that open connections.

> Does audit method keep a map of correlation between a PID and the IP-tuple from the connect() syscall?

Yes: https://github.com/evilsocket/opensnitch/blob/master/daemon/procmon/audit/client.go#L175

gustavo-iniguez-goya:
Actually, we read the events from the daemon auditd, not the kernel audit framework. We could work as that daemon, connecting directly to the kernel to read the events, but there can only be one process connected to the audit framework at a time, and as the auditd daemon comes preinstalled in many distros (suse, redhat, fedora...) we'd probably run into many problems.

themighty1:
Thanks for explainin, it makes sense now.

PR #346, opened 2021-02-13
makefile:

themighty1:
rename, remove unused targets
    trigger targets only when relevant files actually are changed
git:
    do not track files which are generated during make
github actions fix to not rely on a pre-built ui.pb.go


themighty1:
@gustavo-iniguez-goya , any thoughts on this PR ?

gustavo-iniguez-goya:
sorry @themighty1 , I ran into some issues when you pushed it (before 1.3.6 release). As the protocols are no longer on the branch, my scripts to create the packages failed.

But I've tested it again, and just executing `make protocol` before creating the packages resolves the problem.

gustavo-iniguez-goya:
There's an error with the github action: https://github.com/evilsocket/opensnitch/runs/1953642591#step:4:58

Maybe we need to `make protocol`  before getting the dependencies? https://github.com/evilsocket/opensnitch/blob/master/.github/workflows/go.yml#L21

themighty1:
Noted, thanks, will look into it.

PR #354, opened 2021-02-18
do not block connection processing when GUI popup is active.

themighty1:
I tested this by running a script which generated 100 new connections per second and setting the UI popup timeout to 0 with default action deny. The script ran for at least 24h.
The binary was built with -race for race detection.

### DIFF
--- a/daemon/main.go
+++ b/daemon/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"flag"
 	"fmt"
@@ -10,8 +11,8 @@ import (
 	"os/signal"
 	"runtime"
 	"runtime/pprof"
-	"sync"
 	"syscall"
+	"time"
 
 	"github.com/evilsocket/opensnitch/daemon/conman"
 	"github.com/evilsocket/opensnitch/daemon/core"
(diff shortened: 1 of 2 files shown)


gustavo-iniguez-goya:
it works great @themighty1! , then we don't need this timeout anymore:
https://github.com/evilsocket/opensnitch/pull/354/files#diff-dbb5a401e7e78dd50c731b93cf4fb89cea55032c2159782fb590c25a383ff0b5R274

themighty1:
what if the GUI crashes and never responds, we would be stuck waiting forever.

gustavo-iniguez-goya:
well, yes, you're right.

Have you noticed extra errors like this one?
`Error sending packet to queue channel 447358533`

themighty1:
Yes, I see those errors quite often, I feel that the code here
https://github.com/evilsocket/opensnitch/blob/aee92d8d0ee4b72722ea12becdc39db56fcb7274/daemon/netfilter/queue.go#L221-L222

should try sending the packet again and not immediately fall back to the default case .
I will try later to make that code retry sending and see if the error can go away by retrying.

gustavo-iniguez-goya:
ok, no problem. Lets commit this PR and send another one if you want.

PR #368, opened 2021-03-14
fix compilation problem

denysvitali:
For some strange reason (I guess related to the fact that [golang/protobuf switched to google.golang.org/protobuf](https://github.com/golang/protobuf/blame/b5de78c91d0d09482d65f0a96927631cd343d7bb/README.md#L10-L14)) the project, or better the protobuf definitions, were not compilable on my machine.

This PR fixes the problem (for me), but I'm not sure if my configuration is different from all of you: it might not be the best fix.


raboof:
CI failure seems to be #381 (which might conflict with this PR somewhat)

gustavo-iniguez-goya:
Thank you @denysvitali , we finally merge this PR #381  instead.

denysvitali:
Equally awesome, thanks!

PR #375, opened 2021-03-21
[Locale]Add Japanese Translation

gnuhead-chieb:


gustavo-iniguez-goya:
Thank you @gnuhead-chieb !

Could you remove the .qm file from the PR? We generate when generating new packages.

gnuhead-chieb:
removed

Issue #359, opened 2021-02-27
compilation having go issues - note: module requires Go 1.14


NESC1US:
Present yourself (or at least say "Hello" or "Hi") and be kind && respectful.

**Describe the bug**
as mentioned in discussion, compilation instructions are failing for me on Ubuntu 20.04.2 LTS (go1.13.8 linux/amd64) 

**To Reproduce**
follow the [instruction](https://github.com/evilsocket/opensnitch/wiki/Compilation) on Ubuntu 20.04 with Go >= 1.13

1. sudo apt-get install git golang libnetfilter-queue-dev libpcap-dev protobuf-compiler python3-pip pyqt5-dev-tools qttools5-dev-tools qt5-default
2. go get github.com/golang/protobuf/protoc-gen-go; mv go .local
3. export GOPATH=~/.local/go
4. export PATH=$PATH:$GOPATH/bin
5. python3 -m pip install --user grpcio-tools
6. git clone https://github.com/evilsocket/opensnitch
7. cd opensnitch
8. make

**Post error logs:** 
make
make[1]: Entering directory '/home/user/foo/opensnitch/proto'
protoc -I. ui.proto --go_out=plugins=grpc:../daemon/ui/protocol/
python3 -m grpc_tools.protoc -I. --python_out=../ui/opensnitch/ --grpc_python_out=../ui/opensnitch/ ui.proto
make[1]: Leaving directory '/home/user/foo/opensnitch/proto'
make[1]: Entering directory '/home/user/foo/opensnitch/daemon'
go: downloading github.com/evilsocket/ftrace v1.2.0
go: downloading github.com/golang/protobuf v1.0.0
go: downloading github.com/google/gopacket v1.1.14
go: downloading github.com/vishvananda/netlink v1.1.0
go: downloading google.golang.org/grpc v1.11.3
go: extracting github.com/evilsocket/ftrace v1.2.0
go: downloading github.com/fsnotify/fsnotify v1.4.7
go: extracting github.com/vishvananda/netlink v1.1.0
go: downloading golang.org/x/sys v0.0.0-20190606203320-7fc4e5ec1444
go: extracting github.com/google/gopacket v1.1.14
go: extracting github.com/golang/protobuf v1.0.0
go: downloading golang.org/x/net v0.0.0-20180417003750-8d16fa6dc9a8
go: extracting github.com/fsnotify/fsnotify v1.4.7
[...]

Issue #344, opened 2021-02-12
Not works on Python 3.6


bl00mber:
Installer places `grpc` to `/usr/local/lib/python3.6`, but installs `opensnitch` and `PyQt5` to `/usr/lib/python3`, producing #332 error. After `PyQt5` reinstalled to 3.6, if dist_path and version changed to 3.6 in `/usr/bin/opensnitch-ui`, fails with error:
```
Loading translations: /usr/lib/python3/dist-packages/opensnitch/i18n locale: en_US
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
```

gustavo-iniguez-goya:
Hi @bl00mber , what distro are you using? did you installed the GUI manually or using the precompiled packages?

bl00mber:
Ubuntu 16, installed using
`sudo dpkg -i opensnitch_1.3.5-1_amd64.deb python3-opensnitch-ui_1.3.5-1_all.deb; sudo apt -f install`

gustavo-iniguez-goya:
ok, thank you! I'll try to reproduce it.

gustavo-iniguez-goya:
ok, python3.6 does not exist on ubuntu16.04, so please, post all the steps you followed to install python3.6, grpcio, etc, in order to understand what's happening, and to try to reproduce it.

gustavo-iniguez-goya:
It works on Ubuntu 16.04 with python3.5:

`$ sudo dpkg -i python3-opensnitch-ui_1.3.5-1_all.deb; sudo apt -f install`
(select to not install dependencies)
`$ sudo pip3 install grpcio==1.16.1`
`$ sudo pip3 install protobuf`

Notice that I installed the dependencies system wide.

bl00mber:
Thanks! That's how I fixed this:
`python3.5 -m pip install grpcio==1.16.1 protobuf slugify`

gustavo-iniguez-goya:
Great! I've added it to the Documentation/wiki.

By the way, I've realized that latest changes added to master don't work on ubuntu 16.04 (due to some libs versions). I'm investigating it. The effect is that the date of the rules is not updated when a new network event matches a particular rule.

gustavo-iniguez-goya:
Ok, the problem is that UPSERTS (update events if insert fails due to conflicts) weren't introduced until SQLite 3.24.x, and Ubuntu 16.04 SQLite version is 3.11. 

[...]

Issue #332, opened 2021-01-10
GUI not opening on Fedora 32


sebastianelsner:
Hi,

while testing on Fedora 32 I found an issue with the rpms. The gui would not open, erroring with:

```python
Traceback (most recent call last):
  File "/usr/bin/opensnitch-ui", line 16, in <module>
    import grpc
ModuleNotFoundError: No module named 'grpc'
```

I was able to solve it by manually installing

```bash
sudo dnf install python3-grpcio
```

Maybe this should be included in the spec file as a requirement?

Regards

Sebastian

gustavo-iniguez-goya:
Hi Sebastian!

Glad you fixed it by yourself. Yes, the thing is that python3-grpcio is not packaged in all rpm based distributions and versions, and it seems that we're incompatible with some python3-grpcio versions, like the one from OpenSuse Tumbleweed (for whatever reason, [not even a minimal hello-world works](https://stackoverflow.com/questions/65145129/unable-to-get-grpc-server-working-using-python3-grpcio-shipped-with-several-gnu)).

I'm trying to automate the builds to generate packages for Fedora >= 30, with that dependency added.

gustavo-iniguez-goya:
I've released v1.3.5, and I've generated a rpm without the pip dependencies, adding as requirement all the packages that we need for installing it. Tested on f33 and f29, and it works fine :)

I haven't had time to test it on SuSe, but I'll give it a try just in case it works with latest python3-grpcio package.

Thank you for reporting this.

sebastianelsner:
Thank you for fixing!

vbooka1:
> I haven't had time to test it on SuSe, but I'll give it a try just in case it works with latest python3-grpcio package.

just FYI: `pip3 install --user grpcio` does NOT work in openSuSE and results in error "No module named 'grpc'".

installing grpc from repositories `zypper in python3-grpcio` does work.

PR #381, opened 2021-03-28
Support more recent protoc-gen-go

raboof:
When building the project with protoc-gen-go version 1.5.1,
it fails with the following:

```
protoc -I. ui.proto --go_out=plugins=grpc:../daemon/ui/protocol/
protoc-gen-go: unable to determine Go import path for "ui.proto"

Please specify either:
	• a "go_package" option in the .proto source file, or
	• a "M" argument on the command line.

See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information.

--go_out: protoc-gen-go: Plugin failed with status code 1.
```

This can be fixed by adding the full go package as an option in the
proto file. To make sure the code is generated to the correct path,
we also have to add add the `paths=source_relative` option to the
protoc plugin.

After this, the code is generated correctly, but the generated code
references classes like grpc.ClientConnInterface which were introduced
in 1.27.0.

### DIFF
--- a/daemon/go.mod
+++ b/daemon/go.mod
@@ -6,14 +6,14 @@ require (
 	github.com/evilsocket/ftrace v1.2.0
 	github.com/fsnotify/fsnotify v1.4.7
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b // indirect
-	github.com/golang/protobuf v1.0.0
+	github.com/golang/protobuf v1.5.0
 	github.com/google/gopacket v1.1.14
 	github.com/vishvananda/netlink v1.1.0
 	github.com/vishvananda/netns v0.0.0-20191106174202-0a2b9b5464df // indirect
-	golang.org/x/net v0.0.0-20180417003750-8d16fa6dc9a8
+	golang.org/x/net v0.0.0-20190311183353-d8887717615a
 	golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208 // indirect
 	golang.org/x/sys v0.0.0-20190606203320-7fc4e5ec1444 // indirect
 	golang.org/x/text v0.3.0 // indirect
-	google.golang.org/genproto v0.0.0-20180413175816-7fd901a49ba6 // indirect
-	google.golang.org/grpc v1.11.3
+	google.golang.org/grpc v1.27.0
+	google.golang.org/protobuf v1.26.0
 )
--- a/proto/Makefile
+++ b/proto/Makefile
@@ -1,7 +1,7 @@
 all: ../daemon/ui/protocol/ui.pb.go ../ui/opensnitch/ui_pb2.py
 
 ../daemon/ui/protocol/ui.pb.go: ui.proto
-	protoc -I. ui.proto --go_out=plugins=grpc:../daemon/ui/protocol/
+	protoc -I. ui.proto --go_out=plugins=grpc:../daemon/ui/protocol/ --go_opt=paths=source_relative
 
 ../ui/opensnitch/ui_pb2.py: ui.proto
 	python3 -m grpc_tools.protoc -I. --python_out=../ui/opensnitch/ --grpc_python_out=../ui/opensnitch/ ui.proto
--- a/proto/ui.proto
+++ b/proto/ui.proto
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 package protocol;
 
+option go_package = "github.com/evilsocket/opensnitch/daemon/ui/protocol";
+
 service UI {
     rpc Ping(PingRequest) returns (PingReply) {}
     rpc AskRule (Connection) returns (Rule) {}


themighty1:
Thanks, this worked for me. Made the ui/protocol/ui.pb.go:1464:7: undefined: grpc.ClientConnInterface error go away.

themighty1:
@gustavo-iniguez-goya , I think we're gonna need this to get rid of CI builds errors.

gustavo-iniguez-goya:
I was more concerned about the go.mod changes.

until now we've been compatible with the no-dependencies rpm package from >= fedora29. But now that we depend on grpc >= 1.27 maybe we're even incompatible with f32:
`* Thu Mar 25 2021 Benjamin A. Beasley <code@musicinmybrain.net> - 1.26.0-14`
https://koji.fedoraproject.org/koji/buildinfo?buildID=1728134

themighty1:
How about we keep the grpc version as it is now and incorporate this bash script https://github.com/evilsocket/opensnitch/issues/359#issuecomment-787503115
into the Makefile ?

raboof:
> But now that we depend on grpc >= 1.27 maybe we're even incompatible with f32

Good point. If we care about particular distro's perhaps we should try and add them to CI? Do we care about f32? https://fedorapeople.org/groups/schedule/f-34/f-34-key-tasks.html suggests it's going end-of-life next month, so people still using it might be happy to also use an older version of opensnitch?

> https://koji.fedoraproject.org/koji/buildinfo?buildID=1728134

Note I don't see go mentioned in that package. It might be that having a minor version difference between the go library and the python library does not lead to problems? After all, the gRPC wire format is rather stable.

I tried the following:

```
$ docker run -ti fedora:32
# yum install git make protobuf protobuf-devel python3-grpcio golang
# go get google.golang.org/protobuf/cmd/protoc-gen-go \
         google.golang.org/grpc/cmd/protoc-gen-go-grpc
# export PATH="$PATH:$(go env GOPATH)/bin"
# git clone https://github.com/evilsocket/opensnitch
# cd opensnitch
# make
```

This leads to the following error:

```
make[1]: Entering directory '/opensnitch/proto'
protoc -I. ui.proto --go_out=plugins=grpc:../daemon/ui/protocol/ --go_opt=paths=source_relative
--go_out: protoc-gen-go: plugins are not supported; use 'protoc --go-grpc_out=...' to generate gRPC

See https://grpc.io/docs/languages/go/quickstart/#regenerate-grpc-code for more information.
```

I changed the `proto/Makefile` to look like:

```
protoc -I. ui.proto --go-grpc_out=paths=source_relative:../daemon/ui/protocol/
```

Now daemon/ui/protocol/ui_grpc.pb.go seems to be created successfully.

There's another problem with the python code generation, though: `make` now fails with:

```
protoc -I. ui.proto --go-grpc_out=paths=source_relative:../daemon/ui/protocol/
python3 -m grpc_tools.protoc -I. --python_out=../ui/opensnitch/ --grpc_python_out=../ui/opensnitch/ ui.proto
/usr/bin/python3: Error while finding module specification for 'grpc_tools.protoc' (ModuleNotFoundError: No module named 'grpc_tools')
make[1]: *** [Makefile:7: ../ui/opensnitch/ui_pb2.py] Error 1
```

That code seems to have been like that for ages though. Did that ever work on f32?

gustavo-iniguez-goya:
> Good point. If we care about particular distro's perhaps we should try and add them to CI? Do we care about f32? 

actually no, I don't care about f32 in particular, but all the distributions. However the changes you're proposing are for NixOS or you particular setup (from another PR: `My OS comes with grpcio-tools 1.36.1, which seems to work fine`).

> (...) suggests it's going end-of-life next month

people keeps using out-of-date distributions and software, like ubuntu 18.04, ubuntu16 (https://github.com/evilsocket/opensnitch/issues/344), mint19 .. (take a look at the closed issues). Should we drop support for those distributions? I doubt it, as long as we don't require new library/packages versions for fixing issues, etc., or the changes to make it backwards compatible are minimal (https://github.com/evilsocket/opensnitch/commit/3f3e2d6de0cb6287c3e79ce1f2c8dd9a1566a7f5#diff-0e39c69c5a87dd2c07a91a68bea87fd0b9ccf35c1069eb8985d8680de058375b).

> Did that ever work on f32?

yep, in fact from >= f29, https://github.com/evilsocket/opensnitch/issues/332

raboof:
> > (...) suggests it's going end-of-life next month
> 
> people keeps using out-of-date distributions and software, like ubuntu 18.04, ubuntu16 (#344), mint19 .. (take a look at the closed issues). Should we drop support for those distributions? I doubt it, as long as we don't require new library/packages versions for fixing issues, etc., or the changes to make it backwards compatible are minimal ([3f3e2d6#diff-0e39c69c5a87dd2c07a91a68bea87fd0b9ccf35c1069eb8985d8680de058375b](https://github.com/evilsocket/opensnitch/commit/3f3e2d6de0cb6287c3e79ce1f2c8dd9a1566a7f5#diff-0e39c69c5a87dd2c07a91a68bea87fd0b9ccf35c1069eb8985d8680de058375b)).

I completely agree that if we can make things work with both new and old versions that's preferable.

> > Did that ever work on f32?
> 
> yep, in fact from >= f29, #332

That does not prove that `python3 -m grpc_tools.protoc -I. --python_out=../ui/opensnitch/ --grpc_python_out=../ui/opensnitch/ ui.proto` ever worked on f32, though.

That issue is from january, which is before #346. Before #346 we were checking in `ui/opensnitch/ui_pb2.py` so people building opensnitch didn't need to generate it themselves. So it seems we might have broken building on f32 with that change already...

gustavo-iniguez-goya:
> That does not prove that python3 -m grpc_tools.protoc -I. --python_out=../ui/opensnitch/ --grpc_python_out=../ui/opensnitch/ ui.proto ever worked on f32, though.
> That issue is from january, which is before #346. Before #346 we were checking in ui/opensnitch/ui_pb2.py so people building opensnitch didn't need to generate it themselves. So it seems we might have broken building on f32 with that change already...

Well, the only way to prove it is you to checkout the code and see if it builds or not. I never had a problem generating python's grpc code, only Go's. But the fact is it worked for sebastian and me, and now we can't provide a rpm package without pip dependencies.

Another side effect is that existing installations won't have grpcio-tools 1.36.1 in many cases (as I've just tested on my kubuntu 20.04), so we'll need to upgrade it in future versions if we don't want to cause troubles.

By the way, kubuntu 20.04 by default has python3-grpcio 1.16.1 in the repositories, making us incompatible with this change if the user wanted to use that package instead of pip3's. Also now we can't provide deb packages with python3-grpcio dependency for this reason (debian sid: 1.30.2-3, buster: 1.16.1-1, etc). Yes, well, we can, but the number of compatible distros will be minimal as of today.

I'm just listing the issues I'm running into as I test it. I have to take them into account.

raboof:
>> That does not prove that `python3 -m grpc_tools.protoc -I. --python_out=../ui/opensnitch/ --grpc_python_out=../ui/opensnitch/ ui.proto` ever worked on f32, though
>
> Well, the only way to prove it is you to checkout the code and see if it builds or not

It didn't work for me even before merging this PR:

```
$ docker run -ti fedora:32
# yum install git make protobuf protobuf-devel python3-grpcio golang
# go get google.golang.org/protobuf/cmd/protoc-gen-go \
         google.golang.org/grpc/cmd/protoc-gen-go-grpc
# export PATH="$PATH:$(go env GOPATH)/bin"
# git clone https://github.com/evilsocket/opensnitch
# cd opensnitch
# git checkout 148526e^
# cd proto
# python3 -m grpc_tools.protoc -I. --python_out=../ui/opensnitch/ --grpc_python_out=../ui/opensnitch/ ui.proto
/usr/bin/python3: Error while finding module specification for 'grpc_tools.protoc' (ModuleNotFoundError: No module named 'grpc_tools')
```

Did you install grpc_tools on f32? How?

> the fact is it worked for sebastian and me, and now we can't provide a rpm package without pip dependencies.

I agree it doesn't work, but you seem to be saying it was broken by this PR. That doesn't make sense - this PR doesn't touch the python part at all.

> Another side effect is that existing installations won't have grpcio-tools 1.36.1 in many cases (as I've just tested on my kubuntu 20.04), so we'll need to upgrade it in future versions if we don't want to cause troubles
>
> By the way, kubuntu 20.04 by default has python3-grpcio 1.16.1 in the repositories, making us incompatible with this change if the user wanted to use that package instead of pip3's. Also now we can't provide deb packages with python3-grpcio dependency for this reason (debian sid: 1.30.2-3, buster: 1.16.1-1, etc). Yes, well, we can, but the number of compatible distros will be minimal as of today.

I don't understand what this has to do with this PR - this PR only touched the Go generation, not the python part, right?

It sounds like you are talking about #383, is that correct? I'm definitely not proposing we should require grpcio-tools 1.36.1, and that's not what that PR is doing either...

PR #382, opened 2021-03-28
Update to go 1.15

raboof:
As 1.14 is no longer supported since 1.16 has been released.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -7,10 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.13
+    - name: Set up Go 1.15
       uses: actions/setup-go@v1
       with:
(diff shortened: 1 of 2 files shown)


raboof:
CI suggests this needs #381 first

gustavo-iniguez-goya:
https://golang.org/doc/modules/gomod-ref#go
>go minimum-go-version
>minimum-go-version
>    The minimum version of Go required to compile packages in this module. 

Is there any good reason to change this? like some feature that doesn't exist in < 1.15 and makes it to fail?
Unless there's no such reason, we'll not change it. When you compile it it'll use the version you have, not the one in the go.mod.
thank you!

raboof:
> When you compile it it'll use the version you have, not the one in the go.mod

Ok, I did not realize that was 'allowed'. I have no strong reason to update then

PR #383, opened 2021-03-28
Allow using newer grpcio-tools

raboof:
My OS comes with grpcio-tools 1.36.1, which seems to work fine, so it
would be nice to allow the ui to use that instead of building 1.10.1
specifically for opensnitch.

### DIFF
--- a/ui/requirements.txt
+++ b/ui/requirements.txt
@@ -1,4 +1,4 @@
-grpcio-tools==1.10.1
+grpcio-tools>=1.10.1
 pyinotify==0.9.6
 unicode_slugify==0.1.3
 pyqt5>=5.6


raboof:
CI failure is #381

gustavo-iniguez-goya:
One problem I've seen with this approach (requiring latest version), is that in some systems the wheel needs to be compiled (and it takes ages to finish, and sometimes it even fails). I prefer to stick to a given prebuilt version, so the wheel is installed without the need to be compiled (I want to change this on the debian packages, to ease the installation process)

If it's packaged for a particular distribution, then yes, the package should have as dependency whatever version of grpcio-tools is available on the distro (as long as it's compatible). That decission should be the task of the packager.

On the other hand, if you're compiling it for yourself, we should just require the minimum versions we're compatible with, so it's available more widely.

On the other hand

raboof:
Currently opensnitch-ui requires an older version (exactly 1.10.1).

With this PR, it doesn't require the latest version - it just 'loosens' the requirement so that versions later than 1.10.1 (such as 1.36.1) are also allowed.

> we should just require the minimum versions we're compatible with, so it's available more widely.

I think that is what this PR does ;)

gustavo-iniguez-goya:
Hi @raboof ,

This is what happens on Mint 20.1 when grpcio >= 1.36 is used to compile UI's protobuf:
```
Traceback (most recent call last):
  File "/usr/bin/opensnitch-ui", line 22, in <module>
    from opensnitch.service import UIService
  File "/usr/lib/python3/dist-packages/opensnitch/service.py", line 12, in <module>
    import ui_pb2
  File "/usr/lib/python3/dist-packages/opensnitch/ui_pb2.py", line 22, in <module>
    create_key=_descriptor._internal_create_key,
AttributeError: module 'google.protobuf.descriptor' has no attribute '_internal_create_key'
Error in sys.excepthook:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/apport_python_hook.py", line 153, in apport_excepthook
    with os.fdopen(os.open(pr_filename,
FileNotFoundError: [Errno 2] No such file or directory: '/var/crash/_usr_bin_opensnitch-ui.999.crash'

Original exception was:
Traceback (most recent call last):
  File "/usr/bin/opensnitch-ui", line 22, in <module>
    from opensnitch.service import UIService
  File "/usr/lib/python3/dist-packages/opensnitch/service.py", line 12, in <module>
    import ui_pb2
  File "/usr/lib/python3/dist-packages/opensnitch/ui_pb2.py", line 22, in <module>
    create_key=_descriptor._internal_create_key,
AttributeError: module 'google.protobuf.descriptor' has no attribute '_internal_create_key'
```

Any solution to this problem? I'd like you to test it at least on Mint 20.1 and Ubuntu 18.04. 
Using grpcio 1.16.0 to compile the protobuf makes the UI work as expected.

Thank you in advance.

raboof:
It looks like `_internal_create_key` was introduced in version 3.12.0 of both the python runtime and the protoc python plugin. The error above then suggests you were using version >=3.12.0 of the protoc python plugin, but <3.12.0 of the python runtime library.

I tested this with a fresh Mint 20.1 installation. I used the following procedure:

```
$ sudo apt-get update
$ sudo apt-get install git protobuf-compiler golang-goprotobuf-dev python3-grpc-tools golang libnetfilter-queue-dev pyqt5-dev-tools qttools5-dev-tools python3-pip python3-slugify python3-pyinotify qt5-default python3-pyqt5.qtsql python3-grpc-tools
$ git clogit ne https://github.com/evilsocket/opensnitch
$ cd opensnitch
$ git remote add raboof https://github.com/raboof/opensnitch
$ git fetch raboof
$ git checkout allow-newer-grpcio-tools
$ make
$ sudo make install
```

This succeeds and produces an opensnitch-ui that opens for me.

It might be interesting to note that this seems to have used the system protoc, which is version 3.6.1, to generate `ui/opensnitch/ui_pb2.py` (so there's no `_internal_create_key` in there). How you got into a state where the codegen was done with a newer version of the protoc python plugin, but the ui still used an older version of the runtime library, is unclear to me - but it doesn't seem to be caused by this PR.

raboof:
(I did not test this branch with Ubuntu 18.04 yet since because of #403 and #404 even master doesn't work for me on Ubuntu 18.04 right now)

gustavo-iniguez-goya:
> This succeeds and produces an opensnitch-ui that opens for me.

Sorry, I didn't explain myself well. I didn't mean to compile opensnitch on Mint20.1 or ubuntu18.04. That actually doesn't bother me much.

The problem is if I build a deb package on a distro with 1.37.0 (or 1.36, etc), and that package is installed on Mint 20.1, regardless of the grpcio version installed on Mint20.1, the error above is thrown.

If I use grpcio 1.16.0 (or earlier), the deb package can be used on Mint20.1, with any gpcio version (>= 1.10).

raboof:
> The problem is if I build a deb package on a distro with 1.37.0 (or 1.36, etc), and that package is installed on Mint 20.1, regardless of the grpcio version installed on Mint20.1, the error above is thrown.

Ah, I didn't realize there's one `deb` package for all Debian-derived distro's.

So we should keep generating the `ui_pb2.py` using the old generator when creating that `.deb` - but this PR still allows that, right? How do you build the .deb package, and on what distro (version)?

gustavo-iniguez-goya:
Yes, until now we've had a GUI deb package that worked on almost all deb based distros, at least regarding the generated protobuffer.

And yes, I can just install grpcio 1.16.x, generate the protobuffer with this version and distribute it, it's not a problem.

I think that everything is clear now, and this PR can be merged. Thank you @raboof

raboof:
Thanks for the in-depth review!

PR #384, opened 2021-03-28
Fallback to default config when /etc/opensnitchd is not found

raboof:


raboof:
(build failure is #381)

gustavo-iniguez-goya:
What problem did you have to have to hardcode the configuration in code?

raboof:
> What problem did you have to have to hardcode the configuration in code?

The distribution I use, NixOS, discourages 'manually' writing directly to /etc. To quickly try out opensnitch, it is convenient to be able to start opensnitchd without having to create a file in /etc.

(another possible solution in this scenario might be to add a command-line flag to specify where to look for the daemon config - but having the defaults built-in without further parameters would be the most convenient)

gustavo-iniguez-goya:
> The distribution I use, NixOS, discourages 'manually' writing directly to /etc. To quickly try out opensnitch, it is convenient to be able to start opensnitchd without having to create a file in /etc.

How do you do that with apache, postfix, telegraph, cronograph, nginx .. ?

> (another possible solution in this scenario might be to add a command-line flag to specify where to look for the daemon config - but having the defaults built-in without further parameters would be the most convenient)

That makes more sense, add a flag option to specify the directory with rules and configuration files, just like any other software, and default to /etc/opensnitchd/.

raboof:
> > The distribution I use, NixOS, discourages 'manually' writing directly to /etc. To quickly try out opensnitch, it is convenient to be able to start opensnitchd without having to create a file in /etc.
> 
> How do you do that with apache, postfix, telegraph, cronograph, nginx .. ?

You typically install those as 'modules'. Indeed we should create an opensnitch 'module' (https://github.com/NixOS/nixpkgs/issues/118463), but even if we have that it would still be convenient to be able to try out opensnitchd 'ephemerally' without installing the module.

gustavo-iniguez-goya:
Sorry, that's something that has nothing to do with opensnitch. As Debian and others do, make a patch and apply it on your builds. I won't hardcode a configuration in code.

Thank you anyway!

PR #386, opened 2021-03-30
change the python slugify package dependency installation method

z0xyz:
when i used the opensnitch-ui , i was instantly prompted with the pop-up prompting me to either accept / decline the network request .
the problem  arised when i wanted to click allow , the application didn't register the allowance , and the popup remained .
i closed the entire program then i invoked the opensnitch-ui back again , yet this time from the terminal (rather than the desktop environment gui) , and i was prompted back again with the pop-up , and as expected i clicked allow yet the pop-up sit still and nothing happened ,
but this time i checked the terminal as the opensnitch-ui package is now a foreground process at the shell , i noticed this error message 

> NameError: name 'unicode' is not defined
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/opensnitch/dialogs/prompt.py", line 309, in closeEvent
>     self._send_rule()
>   File "/usr/lib/python3/dist-packages/opensnitch/dialogs/prompt.py", line 448, in _send_rule
>     rule_temp_name = self._get_rule_name(self._rule)
>   File "/usr/lib/python3/dist-packages/opensnitch/dialogs/prompt.py", line 424, in _get_rule_name
>     rule_temp_name = slugify("%s %s" % (rule.action, rule.duration))
>   File "/home/sin/.local/lib/python3.6/site-packages/slugify.py", line 24, in slugify
>     unicode(
> NameError: name 'unicode' is not defined

whenever i click allow , nothing happen in the UI application , and this error message gets raised at at the terminal due to the slugify.py file , turned out to be an old package and thus not compatible with python3 updates , i looked up slugify at pypi , and we should pass the new name of the updated slugify package to pip 
, so instead of the command >>sudo apt install python3-slugify ,
we should use the pip package manager instead with >>pip install python-slugify


raboof:
This PR seems to contain a lot of changes not related to the issue you describe (and actually I don't see changes that *do* relate to the issue you describe, but I can have missed them ;) ) - did something go wrong here?

raboof:
> instead of the command `sudo apt install python3-slugify`, we should use the pip package manager instead with `pip install python-slugify`

Are you referring to the wiki here? What distribution are you on? Generally I think when using an apt-based distribution it's better to use the packages than to install dependencies with pip directly, but of course we should make sure the instructions actually work :)

z0xyz:
> Are you referring to the wiki here? 
ya [https://github.com/gustavo-iniguez-goya/opensnitch/wiki/Installation](url) 
> What distribution are you on?
eleementary os hera
, and the debian repositories aren't continually updated with the latest software , rather than that i guess python dependencies are most consistent when using the pip package manager rather than distro-related package manager

gustavo-iniguez-goya:
bear in mind that users still uses distributions that have python2 by default. So the binary `pip` points to pip2, not 3. 
For instance, in a ubuntu 18.04 (which doesn't come with pip by default), just after installing python3-pip, the command pip doesn't exist.

gustavo-iniguez-goya:
on the other hand guys and girls, I'd really like you to open a new issue if you find an error, before submitting a PR. In order to discuss it with other members.

raboof:
> I'd really like you to open a new issue if you find an error, before submitting a PR. In order to discuss it with other members.

Sometimes I find it easier to propose a change as a PR compared to trying to describe it 'abstractly' in an issue. In that case I'm happy to discuss the change on the PR, and make changes or close the PR based on that discussion.

Would you prefer me to first create an issue even in such cases?

gustavo-iniguez-goya:
ah, ok, it's only that I'm not familiarized with that workflow :)
until now people has opened an issue, and based on lack on feedback/answer or on agreements PRs have been proposed.

let's discuss it on the PRs for now.

z0xyz:
> bear in mind that users still use distributions that have python2 by default. So the binary `pip` points to pip2, not 3.
python2 & -nix are inextricable , i assume that i've encountered this problem because i've updated my alternatives long before , and made python3 as the default interpreter , so something has to be done about it because many people do so .

gustavo-iniguez-goya:
reinstall from scratch, do the same steps you did to get the error and post both the logs of the installaion + the error, in order to understand where the error resides. And post also what distro are you using please.

themighty1:
@z0xyz , thanks for you input.
It seems like you attempted to make a change in the wiki, please not that the current wiki is located here:
https://github.com/evilsocket/opensnitch/wiki

At the bottom of each page of the wiki there is a link that explains how to make changes to the wiki.

PR #397, opened 2021-04-03
Use ebpf program to find PID of new connections.

themighty1:
before running the branch you have to compile ebpf_prog/opensnitch.c
    opensnitch.c is an eBPF program. Compilation requires getting kernel source.

    cd opensnitch
    wget https://github.com/torvalds/linux/archive/v5.8.tar.gz
    tar -xf v5.8.tar.gz
    patch linux-5.8/tools/lib/bpf/bpf_helpers.h < ebpf_prog/file.patch
    cp ebpf_prog/opensnitch.c ebpf_prog/Makefile linux-5.8/samples/bpf
    cd linux-5.8 && yes "" | make oldconfig && make prepare && make headers_install # (1 min)
    cd samples/bpf && make
    objdump -h opensnitch.o #you should see many section, number 1 should be called kprobe/tcp_v4_connect
    llvm-strip -g opensnitch.o #remove debug info
    sudo cp opensnitch.o /etc/opensnitchd
    cd ../../../daemon

    --opensnitchd expects to find opensnitch.o in /etc/opensnitchd/
    --start opensnitchd with:

    opensnitchd -rules-path /etc/opensnitchd/rules -process-monitor-method ebpf

### DIFF
--- a/daemon/conman/connection.go
+++ b/daemon/conman/connection.go
@@ -13,6 +13,7 @@ import (
 	"github.com/evilsocket/opensnitch/daemon/netlink"
 	"github.com/evilsocket/opensnitch/daemon/netstat"
 	"github.com/evilsocket/opensnitch/daemon/procmon"
+	"github.com/evilsocket/opensnitch/daemon/procmon/ebpf"
 	"github.com/evilsocket/opensnitch/daemon/ui/protocol"
 
 	"github.com/google/gopacket/layers"
@@ -81,44 +82,52 @@ func newConnectionImpl(nfp *netfilter.Packet, c *Connection, protoType string) (
 		INode:   -1,
 	}
 
-	// 0. lookup uid and inode via netlink. Can return several inodes.
-	// 1. lookup uid and inode using /proc/net/(udp|tcp|udplite)
-	// 2. lookup pid by inode
-	// 3. if this is coming from us, just accept
-	// 4. lookup process info by pid
-	uid, inodeList := netlink.GetSocketInfo(c.Protocol, c.SrcIP, c.SrcPort, c.DstIP, c.DstPort)
-	if len(inodeList) == 0 {
-		if c.Entry = netstat.FindEntry(c.Protocol, c.SrcIP, c.SrcPort, c.DstIP, c.DstPort); c.Entry == nil {
-			return nil, fmt.Errorf("Could not find netstat entry for: %s", c)
+	var pid int
+	if procmon.MethodIsEbpf() {
+		pid, err = ebpf.GetPid(c.Protocol, c.SrcPort, c.SrcIP, c.DstIP, c.DstPort)
+		if err != nil {
+			return nil, nil
 		}
-		if c.Entry.INode > 0 {
-			log.Debug("connection found in netstat: %v", c.Entry)
-			inodeList = append([]int{c.Entry.INode}, inodeList...)
+	} else {
+		// 0. lookup uid and inode via netlink. Can return several inodes.
+		// 1. lookup uid and inode using /proc/net/(udp|tcp|udplite)
+		// 2. lookup pid by inode
+		// 3. if this is coming from us, just accept
+		// 4. lookup process info by pid
+		uid, inodeList := netlink.GetSocketInfo(c.Protocol, c.SrcIP, c.SrcPort, c.DstIP, c.DstPort)
+		if len(inodeList) == 0 {
+			if c.Entry = netstat.FindEntry(c.Protocol, c.SrcIP, c.SrcPort, c.DstIP, c.DstPort); c.Entry == nil {
+				return nil, fmt.Errorf("Could not find netstat entry for: %s", c)
+			}
+			if c.Entry.INode > 0 {
+				log.Debug("connection found in netstat: %v", c.Entry)
+				inodeList = append([]int{c.Entry.INode}, inodeList...)
+			}
 		}
-	}
-	if len(inodeList) == 0 {
-		log.Debug("<== no inodes found, applying default action.")
-		return nil, nil
-	}
+		if len(inodeList) == 0 {
+			log.Debug("<== no inodes found, applying default action.")
+			return nil, nil
+		}
 
-	if uid != -1 {
-		c.Entry.UserId = uid
-	} else if c.Entry.UserId == -1 && nfp.UID != 0xffffffff {
-		c.Entry.UserId = int(nfp.UID)
-	}
-
-	pid := -1
-	for n, inode := range inodeList {
-		if pid = procmon.GetPIDFromINode(inode, fmt.Sprint(inode, c.SrcIP, c.SrcPort, c.DstIP, c.DstPort)); pid == os.Getpid() {
-			// return a Process object with our PID, to be able to exclude our own connections
-			// (to the UI on a local socket for example)
-			c.Process = procmon.NewProcess(pid, "")
-			return c, nil
+		if uid != -1 {
+			c.Entry.UserId = uid
+		} else if c.Entry.UserId == -1 && nfp.UID != 0xffffffff {
+			c.Entry.UserId = int(nfp.UID)
 		}
-		if pid != -1 {
-			log.Debug("[%d] PID found %d", n, pid)
-			c.Entry.INode = inode
-			break
+
+		pid = -1
+		for n, inode := range inodeList {
+			if pid = procmon.GetPIDFromINode(inode, fmt.Sprint(inode, c.SrcIP, c.SrcPort, c.DstIP, c.DstPort)); pid == os.Getpid() {
+				// return a Process object with our PID, to be able to exclude our own connections
+				// (to the UI on a local socket for example)
+				c.Process = procmon.NewProcess(pid, "")
+				return c, nil
+			}
+			if pid != -1 {
+				log.Debug("[%d] PID found %d", n, pid)
+				c.Entry.INode = inode
+				break
+			}
 		}
 	}
 	if c.Process = procmon.FindProcess(pid, showUnknownCons); c.Process == nil {
--- a/daemon/conman/connection.go
+++ b/daemon/conman/connection.go
@@ -82,14 +82,18 @@ func newConnectionImpl(nfp *netfilter.Packet, c *Connection, protoType string) (
 		INode:   -1,
 	}
 
-	var pid int
+	pid := -1
 	var uid int
 	if procmon.MethodIsEbpf() {
 		pid, uid, err = ebpf.GetPid(c.Protocol, c.SrcPort, c.SrcIP, c.DstIP, c.DstPort)
 		if err != nil {
+			log.Warning("ebpf warning: %v", err)
 			return nil, nil
 		}
-	} else {
+	} 
+	// sometimes when using eBPF the connection is not found, but falling back to legacy
+	// methods helps to find it and avoid "unknown/kernel pop-ups". TODO: investigate
+	if pid == -1 {
 		// 0. lookup uid and inode via netlink. Can return several inodes.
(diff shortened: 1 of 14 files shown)


themighty1:
I had to separate procmon's methods Init() and End() into a separate package "monitor" in order to avoid cyclic dependencies.

I hard-coded "-100" as the pid for the kernel and the name of process "Linux kernel". Feel free to change it for something better.

Like discussed earlier, in connection.go I did
 if procmon.MethodIsEbpf() { } 
else { the previous code path without modifications }

gustavo-iniguez-goya:
suerb @themighty1 

I'll test it as much as I can and report you the errors I encounter before commiting it.

Having the proc method configured, launch it and change the method to ebpf in the default_config.json:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x686d55]

goroutine 53 [running]:
github.com/evilsocket/opensnitch/daemon/procmon/ebpf.getPidFromEbpf(0xa3ebb1, 0x3, 0xa058, 0xc005b5678c, 0x4, 0x30, 0xc005b56790, 0x4, 0x2c, 0x235e, ...)
	/tmp/381/opensnitch/daemon/procmon/ebpf/ebpf.go:280 +0x175
github.com/evilsocket/opensnitch/daemon/procmon/ebpf.GetPid(0xa3ebb1, 0x3, 0xa058, 0xc005b5678c, 0x4, 0x30, 0xc005b56790, 0x4, 0x2c, 0x235e, ...)
	/tmp/381/opensnitch/daemon/procmon/ebpf/ebpf.go:222 +0xb5
github.com/evilsocket/opensnitch/daemon/conman.newConnectionImpl(0xc004aba1e0, 0xc005f34280, 0x0, 0x0, 0x0, 0x0, 0x0)
	/tmp/381/opensnitch/daemon/conman/connection.go:87 +0x394
github.com/evilsocket/opensnitch/daemon/conman.NewConnection(0xc004aba1e0, 0xc004aba1e0, 0x4662a0, 0xc001581e50)
	/tmp/381/opensnitch/daemon/conman/connection.go:157 +0x14a

```

gustavo-iniguez-goya:
we need to get the uid of the connections, some rules are specific for one user and it keeps show pop-ups to allow/deny the connection.
On the one hand from ebpf using `bpf_get_current_uid_gid();` and from the already established connections ((sock*).UID): 
https://github.com/evilsocket/opensnitch/pull/397/files#diff-60e9326c73b9fc44e79af7900c9979641b1f38043bd532abee4d7b25deca801dR315

On the other hand, this block should go after L132:
https://github.com/evilsocket/opensnitch/pull/397/files#diff-15d7687250ed7ebda3aa02642d66c977276676567dc914a625eb47f5185fcfbcR112-R115

that way we can use it with any monitor methods.

themighty1:
Ah, yes, good catch. I will implement and add to this PR.

daemon/conman/connection.go
```diff
@@ -107,26 +128,18 @@ func newConnectionImpl(nfp *netfilter.Packet, c *Connection, protoType string) (
 		c.Entry.UserId = int(nfp.UID)
```
themighty1, line 128: @gustavo-iniguez-goya , do you know what this line is for
> else if c.Entry.UserId == -1 && nfp.UID != 0xffffffff

more specifically, why can we not rely on UID reported from netfilter?

themighty1:
This PR is ready with the new change. Don't forget to recompile opensnitch.o

daemon/conman/connection.go
gustavo-iniguez-goya: in some systems libnetfilter_queue doesn't obtain the UID (<= 1.0.3 if I remember correctly).

gustavo-iniguez-goya: moreover, some connections inititated from a container doesn't report it:
new connection udp => 41809:172.17.0.5 -> 208.67.220.220:53 uid: 4294967295

probably because going through the FORWARD table.

themighty1: ok, I'll add it as a comment to the code then.

themighty1: So, is it fair to say that what the code does is - if it fails to find the UID, it falls back to using the UID reported via netfilter_queue, even though that UID potentially may be incorrect ?

ebpf_prog/opensnitch.c
```diff
@@ -0,0 +1,446 @@
...
+	u32 zero_key = 0;
+	u64 *counterVal = bpf_map_lookup_elem(&udpv6counter, &zero_key);
+	if (counterVal == NULL){return 0;}
+	struct udpv6_value_t *lookedupValue = bpf_map_lookup_elem(&udpv6Map, &udpv6_key);
+	u64 pid = bpf_get_current_pid_tgid() >> 32;
+	if ( lookedupValue == NULL || lookedupValue->pid != pid) {
+		struct udpv6_value_t udpv6_value;
+        __builtin_memset(&udpv6_value, 0, sizeof(udpv6_value));
+		udpv6_value.pid = pid;
+		udpv6_value.uid = bpf_get_current_uid_gid() >> 32;
```
gustavo-iniguez-goya, line 419: for TCP connections it seems to work fine, but for UDP connections it doesn't obtain the UID. It's either empty or a random int.

themighty1: I println my uid on udp connections while torrenting, I only see uid 1000, nothing is off for me.

gustavo-iniguez-goya:
There're some connections that are not found via ebpf nor in alreadyEstablished connections, but that are already cached:

```
[2021-04-04 11:39:12]  DBG  [0/1] outgoing connection: 57922:192.168.1.103 -> 140.82.121.3:443 || netlink response: 57922:192.168.1.103 -> 140.82.121.3:443 inode: 1563524 - loopback: false multicast: false unspecified: false linklocalunicast: false ifaceLocalMulticast: false GlobalUni: true 
*** pid -1 from ebpf, trying other methods
[2021-04-04 11:39:12]  DBG  [0/1] outgoing connection: 57922:192.168.1.103 -> 140.82.121.3:443 || netlink response: 57922:192.168.1.103 -> 140.82.121.3:443 inode: 1563524 - loopback: false multicast: false unspecified: false linklocalunicast: false ifaceLocalMulticast: false GlobalUni: true 
[2021-04-04 11:39:12]  DBG  Socket found in known pids 2.264491ms, pid: 7946, inode: 1563524, pos: 12, pids in cache: 14
[2021-04-04 11:39:12]  DBG  [0] PID found 7946
[2021-04-04 11:39:12]  DBG  ✔ /usr/lib/firefox-esr/firefox-esr -> 140.82.121.3:443 (002-allow-firefox-common-ports)

```

which makes me wonder if we need `findInAlreadyEstablishedTCP()`
if we do this in `conman/connection.go` :
```
pid := -1
if procmon.MethodIsEbpf() {
(...) pid, uid := ...
}

if pid == -1{
   fmt.Println("*** pid -1 from ebpf, trying other methods")
   /* get inode from netlink *
   /* get inode from netstat*/
   /* get pid from cache, etc*/
}
```

themighty1:
I honestly have never seen a "non-esoteric" process like firefox not being picked up by ebpf.
I made changes to my local code like you did to try the proc method if ebpf method returns -1 and to os.Exit(1) if proc method worked after ebpf method didnt.
Let's see if anything comes up.

daemon/conman/connection.go
gustavo-iniguez-goya: yes, I think so. We'll run into this scenario if we add a FORWARD rule to intercept containerized connections (it's working fine as far as I can tell). In the near future would be great if we could display the container name.
But if an external connection to the box traverses the FORWARD table then we won't get the UID.

gustavo-iniguez-goya:
maybe it's due to the docker iptables rules of my system, even if I stop the docker daemon the rules are not flushed.
I've flushed them all, and it seems to work much better now,

I can confirm that we don't miss any transmission connection anymore :tada: which is a very good indicator.

On the other hand, adding a FORWARD rule to intercept containerized connections works, but it'll need a bit more of work.

gustavo-iniguez-goya:
bpf .o object tested successfully on kernels 4.15.0, 5.4.0, 5.8.4 and 5.10.0

The problem getting the UID on UDP connections is reproducible on ubuntu 18.04 with kernels 4.15 or 5.4. 

And the segfault that I saw before is reproducible following these steps:
- set in default-config.json ProcMonitorMethod to proc
- launch the daemon
- change ProcMonitorMethod to ebpf

```
github.com/evilsocket/opensnitch/daemon/procmon/ebpf.getPidFromEbpf(0xb1112c, 0x3, 0xab48, 0xc000122f4c, 0x4, 0x30, 0xc000122f50, 0x4, 0x2c, 0x17, ...)
        /home/test/397-ebpf/opensnitch/daemon/procmon/ebpf/ebpf.go:280 +0x160
github.com/evilsocket/opensnitch/daemon/procmon/ebpf.GetPid(0xb1112c, 0x3, 0xab48, 0xc000122f4c, 0x4, 0x30, 0xc000122f50, 0x4, 0x2c, 0x17, ...)
        /home/test/397-ebpf/opensnitch/daemon/procmon/ebpf/ebpf.go:221 +0xb5
github.com/evilsocket/opensnitch/daemon/conman.newConnectionImpl(0xc000a83170, 0xc00012d180, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/test/397-ebpf/opensnitch/daemon/conman/connection.go:88 +0x390
github.com/evilsocket/opensnitch/daemon/conman.NewConnection(0xc000a83170, 0xc000a83170, 0x206d3030315b1b6d, 0x6d305b1b207325)
        /home/test/397-ebpf/opensnitch/daemon/conman/connection.go:162 +0x14a
```

gustavo-iniguez-goya:
> The problem getting the UID on UDP connections is reproducible on ubuntu 18.04 with kernels 4.15 or 5.4.

This problem is fixed by using: `bpf_get_current_uid_gid() & 0xffffffff` instead of >> 32.

gustavo-iniguez-goya:
tested on ubuntu 14, kernel 4.4.0 (more due to the kernel than the distro)

ebpf interception works, but dumping active connections from kernel (` daemonNetlink.SocketsDump`) fails despite of the kmods udp_diag, tcp_diag and inet_diag being loaded. Due to this, ebpf.Start() fails and fallbacks to proc.

The other method to get connections from kernel (`netlinkg.GetSocketInfo`) works fine. So maybe instead of [returning here](https://github.com/evilsocket/opensnitch/pull/397/files#diff-60e9326c73b9fc44e79af7900c9979641b1f38043bd532abee4d7b25deca801dR108), we could just continue without monitoring the active connections. But just maybe, I don't see it very important.

themighty1:
@gustavo-iniguez-goya , well done on the research! Could you please take over this PR and modify it as you see fit and pull it. I'm gonna have a busy week, so I will not be able to contribute much over the next few days. Thanks.

gustavo-iniguez-goya:
Sure, it's just a few minimal changes left.

gustavo-iniguez-goya:
note: at least one of the **Linux kernel** connections was a udp6 connection made by dhclient to 547 port.
it can be forced by restarting network-manager on a ubuntu18.04 (service network-manager restart), but the "ghost" connection maybe has been caused by an external dhcp broadcast to reload/change configuration.

gustavo-iniguez-goya:
for the record: the segfault when changing the configuration by hand (from `proc` to `ebpf`) is because when the new monitoring method is set, we don't stop the existing one and start the new one, so the object `hostByOrder` is nil. That process is only done when the config is changed from the GUI. We'll also need to handle it if the config is changed by hand (like if you're in a server and you want to change it by hand).

Issue #394, opened 2021-04-02
Browser based translation/localization


ovari:
Can instructions to the [README.md](https://github.com/evilsocket/opensnitch/tree/master/ui/i18n#readme) on how to provide translations using a browser please be added?

We would like to provide a `hu` (Hungarian - magyar) translation.

Thank you

P.S. We heard about your program from [Snitching on Phones That Snitch On You](https://puri.sm/posts/snitching-on-phones-that-snitch-on-you/)

> OpenSnitch is a really powerful tool but software like this requires a lot of time spent training the firewall, and can sometimes cause odd app errors until you realize the firewall is just doing it’s job. It would definitely benefit from a set of “known good” baseline rules you could apply so you only get prompted for the real outliers. Because of this I don’t know that it’s something the average user would want to install by default, but it’s definitely something useful for people facing more extreme threats.
>
> This would also be a great tool for an IT organization to deploy throughout a fleet of computers along with custom rules that factor in their known good services. It would add an additional layer of protection that would be relatively seamless for their employees.

gustavo-iniguez-goya:
hi @ovari ,

You can read the instructions here: https://github.com/evilsocket/opensnitch/tree/master/ui/i18n

If you have any problem don't hesitate to ask.

Danny3:
I would like to contribute too with translations for my language, but with my limited time I hope that one day I could use something simpler like Transifex where the are 90% of the other projects I contribute to and which already uses the Gettext (.po) system. 
As far as I remember Transifex it's free for open source programs.

gustavo-iniguez-goya:
I can help with the translations. Tell me what your locale code is (hu_HU, fi_FI, etc) and I'll create the files for you to translate using `linguist`. Post the translations back here again, I'll test them, and if everything is fine I'll commit them.

ovari:
Previously the locale code was `hu_HU`; however, the locale code is now `hu`.

A list of locales are shown in the ... drop-down at:
https://transvision.mozfr.org/

Do your other locales need updating to remove the capitals, the country, label?

Thank you

Danny3:
I need for Romanian
I assume the code should be ro_RO

gustavo-iniguez-goya:
> Do your other locales need updating to remove the capitals, the country, label?

As far as I can tell no. On Debian based systems there's no way to select a locale without the country, according to `dpkg-reconfigure locales`. You can also check out what's in this dir `~$ ls /usr/share/i18n/locales`

[...]

PR #418, opened 2021-05-20
Update opensnitch-hu_HU.ts

ovari:
Do `<translation>   …   </translation>` need to have leading and trailing spaces? https://github.com/evilsocket/opensnitch/issues/394#issuecomment-840848423

https://github.com/evilsocket/opensnitch/issues/394

### DIFF
--- a/ui/i18n/locales/hu_HU/opensnitch-hu_HU.ts
+++ b/ui/i18n/locales/hu_HU/opensnitch-hu_HU.ts
@@ -404,7 +404,7 @@
     <message>
         <location filename="../../../opensnitch/res/preferences.ui" line="536"/>
         <source>audit</source>
-        <translation>naplózás</translation>
+        <translation>audit</translation>
     </message>
     <message>
         <location filename="../../../opensnitch/res/preferences.ui" line="541"/>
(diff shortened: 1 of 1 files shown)


ui/i18n/locales/hu_HU/opensnitch-hu_HU.ts
```diff
@@ -6,656 +6,656 @@
...
         <location filename="../../../opensnitch/res/preferences.ui" line="531"/>
         <source>ebpf</source>
-        <translation type="unfinished"></translation>
+        <translation>ebpf</translation>
     </message>
     <message>
         <location filename="../../../opensnitch/res/preferences.ui" line="536"/>
         <source>audit</source>
-        <translation type="unfinished"></translation>
+        <translation>naplózás</translation>
```
gustavo-iniguez-goya, line 407: don't translate this string, it's the name of the process monitor method (auditd).

Sorry for that, I'll make it not translatable for future releases.

gustavo-iniguez-goya:
> Do <translation> … </translation> need to have leading and trailing spaces?

It depends on the string, but usually no.

ui/i18n/locales/hu_HU/opensnitch-hu_HU.ts
ovari: done

ovari:
> > Do  …  need to have leading and trailing spaces?
> 
> It depends on the string, but usually no.

Removed leading and trailing spaces in the translation strings.

gustavo-iniguez-goya:
Thank you very much @ovari !

ovari:
1. Is `asocciated` spelt incorrectly in the en source strings? Should it be spelt as `associated `?
2. Should `opensnitch` be spelt as `OpenSnitch`?
3. Should `pop-up`, `popup` be standardized to be all with or without a hyphen?
https://www.microsoft.com/en-us/language/Search?&searchTerm=popup&langID=382&Source=true&productid=0
4. Should the leading and trailing spaces in the en source strings be removed?

````
<message>
        <location filename="../../../opensnitch/res/preferences.ui" line="578"/>
        <source>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;If checked, opensnitch will prompt you to allow or deny connections that don&apos;t have an asocciated PID, due to several reasons.&lt;/p&gt;&lt;p&gt;The pop-up dialog will only contain information about the network connection.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</source>
        <translation></translation>
    </message>
````

Thank you

ovari:
@gustavo-iniguez-goya please contact us in the future if there are new strings to translate.

Thank you once again.

PR #486, opened 2021-08-30
Update README.md

ovari:


gustavo-iniguez-goya:
Hi @ovari ,

No, translations won't be done using weblate. Those of you who prefer to use weblate are free to use weblate or other tools, but we'll stick to the instructions here: https://github.com/evilsocket/opensnitch/tree/master/ui/i18n

I'd prefer to update that README instead, stating that weblate can be used.

README.md
```diff
@@ -18,6 +18,12 @@
 
 Please, refer to [the documentation](https://github.com/evilsocket/opensnitch/wiki) for detailed information.
 
+### Translations
```
gustavo-iniguez-goya, line 21: @ovari how about this:
```
You can help translating OpenSnitch to your language following these instructions:
https://github.com/evilsocket/opensnitch/tree/master/ui/i18n

If you wish you can use https://hosted.weblate.org/projects/opensnitch/gui/ for this task.
Thank you so much for your effort in helping us keep it available in many languages!
```

The reason for this is that I expect translators to send PRs with new/updated languges, and the translations be tested. That's why using `linguist` comes in handy, you can view in real time if a sentence breaks the layout or not.

Otherwise these tasks will end up on the maintainers' side.

ovari: Sorry, don't know where this went.

Can this be deleted now or wait for the other merge request to be accepted.

Hopefully the merge request is want you want now.

Thank you

PR #504, opened 2021-09-14
go: module github.com/golang/protobuf is deprecated: Use the "google.…

themighty1:
…golang.org/protobuf" module instead.


gustavo-iniguez-goya:
correct, thank you!

Issue #502, opened 2021-09-13
Can't connect to NFS share while enabled


Natrinicle:
Hi there, love the project so far but I seem to have hit a bug that's a show stopper for me.

While OpenSnitch is enabled I'm unable to connect to my NFSv4 server. If I have a previously open connection before I start OpenSnitch it seems to mark it as related and doesn't block it but if I start OpenSnitch before connecting it's unable and times out. This is despite having a rule for all traffic to that host as allowed.

1. Create an NFSv4 server on the network, mount this NFS share locally.

2. This is my sanitized /etc/systemd/system/mnt-tank.mount that I use to mount on my local machine.
```systemd
[Unit]
Description=<hostname>:/mnt/tank nfs mount script
After=network-online.service
Wants=network-online.service

[Mount]
What=<hostname>:/mnt/tank
Where=/mnt/tank
Options=rw,vers=4.2,rsize=1048576,wsize=1048576,timeo=3,retrans=5
Type=nfs
TimeoutSec=30

[Install]
WantedBy=multi-user.target
```

3. The OpenSnitch rule is allow to <hostname> forever.

4. To bring up the mount, run `systemctl start mnt-tank.mount`

**OS (please complete the following information):**
 - OS: Debian GNU/Linux
 - Window Manager: KDE
 - Kernel version: 5.10.0-8-amd64
 - Version: Testing (Bookworm)

Client & Server:
nfs-kernel-server: 1:1.3.4-6

gustavo-iniguez-goya:
Thank you for reporting this issue @Natrinicle , I'll try to reproduce it.

Do you have the option `Debug invalid connections` checked under `Preferences->Nodes`? If not, check it please and let me know if opensnitch prompts you to allow or deny an "Outgoing connection".

Natrinicle:
I did not originally, went in and turned `Debug invalid connections` on, unmounted the NFS share with OpenSnitch disabled, then remounted with OpenSnitch enabled and there are no additional connection prompts nor denied events in the events tab.

gustavo-iniguez-goya:
Ok, thank you.

I've got good news and bad news:

- The good news:

The change #493 to intercept WireGuard connections #454  also allow us to intercept NFS connections (as well as ip tunnels #500):

![image](https://user-images.githubusercontent.com/2742953/133145448-00910d73-e6d0-4a1b-b417-bc5d5f48d4de.png)

Once allowed the share is mounted properly.

- The bad news:

[...]

Issue #500, opened 2021-09-11
Opensnitch doesn't let through ICMP ECHO packets if the UI is running


markozajc:
Greetings! I've encountered a bug in opensnitch that can cause numerious issues, including filesystem hangs for NFS and 
generally broken behaviour in anything networked. I have pinpointed it to opensnitch not letting through outbound ICMP ECHO packets, but it only occurs if the UI is running.

**Describe the bug**
Opensnitch 1.4.0 blocks all ICMP ECHO packets when the UI is running. They do not show up on the log.

**To Reproduce**
1. Install opensnitch 1.4.0
2. Start opensnitchd
3. Run opensnitch-ui
4. Attempt to ping any host

**Post error logs:** 
N/A

**Expected behavior (optional)**
Outbound ICMP ECHO packets aren't filtered.

**Screenshots**
N/A

**OS (please complete the following information):**
 - OS: Bedrock Linux, opensnitch installed on the Debian stratum
 - Window Manager: KWin
 - Kernel version: `Linux [hostname] 5.13.0-15.1-liquorix-amd64 #1 ZEN SMP PREEMPT liquorix 5.13-13.1~bookworm (2021-09-08) x86_64 GNU/Linux`
 - Version Debian 12/Bookworm

**Additional context**
This bug was not present in 1.3.6, but stopped working in one of the 1.4.0 RC's (I think it was .3, tell me if you want me to pinpoint it to a specific version).

markozajc:
My mistake, for some reason `/etc/opensnitchd/system-fw.json` wasn't installed.

gustavo-iniguez-goya:
:+1: :wink:

markozajc:
@gustavo-iniguez-goya Actually, I'm getting the same issue for my IPv6 tunnel. Is there a known fix or should I reopen the issue.

gustavo-iniguez-goya:
what's that issue about? ipv6 tunnel == VPN?

markozajc:
The issue is really the same as previously; ICMP ECHO requests getting dropped, but this time for IPv6 only. It's not a VPN tunnel, it's a v4tunnel, with Tunnel Broker as the endpoint provider.

gustavo-iniguez-goya:
yay, I see. Please, add this block to the system-fw.json file:
```
"Rule": {
                "Description": "Allow icmp",
                "Table": "mangle",
                "Chain": "OUTPUT",
                "Parameters": "-p ipv6-icmp",
[...]

Issue #454, opened 2021-07-10
VPN doesn't work with opensnitchd most of the time [wireguard, iptables]


chicagoo17:
Hi there, only started my journey in linux and maybe just don't understand something.

My wireguard vpn doesn't work **sometimes** _(sometimes if I turn on vpn after opensnitchd it works, around 30% cases)_ when opensnitchd runs.

That seems to be because of an iptables rule:
**-A INPUT -p udp -m udp --sport 53 -j NFQUEUE --queue-num 0 --queue-bypass**

If I delete this rule everything works fine, but I didn't manage to fully understand what it does.
If it matters, my wireguard .conf uses  "DNS = 10.0.0.1" (and IPv6 address).
I don't quite understand where to see logs of wireguard or opensnitch for this (I couldn't find this) so just asking.

Questions:
**1. What is the purpose of this iptables rule above?**
**2. Is it safe for my safety and privacy to delete this rule?**
_3. Do I need to configure my vpn to bypass this?_
_4. By the way what do these lines mean? It's from /var/log/opensnitchd.log_
_Error while pinging UI service: rpc error: code = DeadlineExceeded desc = context deadline exceeded, state: READY_
_[date and time]  ERR  getting notifications: rpc error: code = Unavailable desc = transport is closing <nil>_
_[date and time]]  WAR  Error while pinging UI service: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial unix /tmp/osui.sock: connect: connection refused", state: TRANSIENT_FAILURE_

chicagoo17:
@gustavo-iniguez-goya Could help me please?

gustavo-iniguez-goya:
Hi @chicagoo17 !

First things firsts, could you tell us the opensnitch version and GNU/Linux distribution you're using?


> 1. What is the purpose of this iptables rule above?

To intercept DNS responses, so we can keep a cache of known IP-host list. If a new process tries to establish a connection to an IP and it's in the cache, we display the host instead of the IP (which is anyway shown on the popups).

> 2. Is it safe for my safety and privacy to delete this rule?

Yes, you can delete the rule, it won't affect the interception of outbound connections. But we check every 30s if that rule is added, and if it's not present we add it again (it's pretty common to disable fw rules once you get access to a system).

> 3. Do I need to configure my vpn to bypass this?

Actually, this is the first time I see this problem. 

> 4. By the way what do these lines mean? It's from /var/log/opensnitchd.log
[...]

PR #513, opened 2021-09-18
Allow to intercept more kernel connections

gustavo-iniguez-goya:
Some connections are initiated from kernel space, like WireGuard VPNs (#454), NFS or SMB connections (#502) and ip tunnels (#500).

More information regarding this change: #493

TODO:
- support for i386: there's an invalid access reading udp header.
- test it on arm*.

### DIFF
--- a/daemon/procmon/details.go
+++ b/daemon/procmon/details.go
@@ -41,7 +41,7 @@ func (p *Process) readComm() error {
 	if err != nil {
 		return err
 	}
-	p.Comm = string(data)
+	p.Comm = core.Trim(string(data))
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


gustavo-iniguez-goya:
Tested on aarch64, kernel 5.8:
- :white_check_mark: NFS and CIFS (smb) interception works.
- :x: VPN WireGuard interception works, **but** source and destination ports are always 0. If instead of read them from UDP header we read them from sock struct, the source port is not 0 but destination port is still 0:
     ```
     if (udp_key.sport == 0){
                bpf_probe_read(&udp_key.sport, sizeof(udp_key.sport), &sk->__sk_common.skc_num);
     }
     if (udp_key.dport == 0){
                bpf_probe_read(&udp_key.dport, sizeof(udp_key.dport), &sk->__sk_common.skc_dport);
     }

     ```

    However tracing these connections with bpftrace report both ports just fine.

PR #574, opened 2021-12-23
Update block-lists.md

Boruch-Baum:
Correct dead link. Update text to reflect change in target file-name


gustavo-iniguez-goya:
Thank you !

PR #586, opened 2022-01-05
add Repology badge to README

Scrumplex:
This banner can help people find their corresponding distro packages.


gustavo-iniguez-goya:
Hi @Scrumplex !

it seems to break the style?

image

What if we put it below the other badges?

image

Scrumplex:
It makes the README arbitrarily long if we just put it below the other badges. If we had some text in the README we could put it on the right side and let the text wrap around it

Scrumplex:
Maybe we can also put this in the wiki instead of the README?

gustavo-iniguez-goya:
Absolutely @Scrumplex , here's how: https://github.com/evilsocket/opensnitch/blob/wiki/README.md

It fits perfectly on the wiki home page: https://github.com/evilsocket/opensnitch/wiki

It's a pity that Repology don't offer an horizontal badge of all repositories.

gustavo-iniguez-goya:
Now that the badge has been added to the wiki I think this PR can be closed. Or in case to be added to the README, since everything is center aligned, it must be center aligned in order not to break the current style.

Scrumplex:
Oh right. I updated this PR to just add a badge instead of the full list. See https://github.com/evilsocket/opensnitch/blob/8d54ca637d5a82ff193115ee96322679967663d4/README.md

gustavo-iniguez-goya:
Thank you!

PR #608, opened 2022-02-02
Improve systemd service unit name

staticssleever668:
It was a full sentence that looked out of place. Official documentation
of systemd recommends to make it a short capitalized label, preferably a
noun.
See [man page `systemd.unit`](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Description=).


gustavo-iniguez-goya:
thank you!

PR #610, opened 2022-02-05
Fix a minor typo

markozajc:
This PR fixes the following typo:
`Disable pop-ups, only display an notification` to `Disable pop-ups, only display a notification`


gustavo-iniguez-goya:
Thank you @markozajc ! it also needs to be changed here: https://github.com/evilsocket/opensnitch/blob/c396ed652c693d2ccccf514473044e456c157608/ui/opensnitch/res/preferences.ui#L461

markozajc:
I've changed it already in 65bd9153dbaf4b5cf6b883d275cc5bcda00531d1, but you're viewing the wrong commit (c396ed652c693d2ccccf514473044e456c157608).

gustavo-iniguez-goya:
you're right O:) thank you!

PR #616, opened 2022-02-09
feat: add protobuf generated files

denysvitali:
If you want to use the UI module in another project, Go will not
allow you to import the package because the ui protobuf package
is missing. We have to commit it so that we can allow this
project to be used as a library.


gustavo-iniguez-goya:
hey @denysvitali , they were added to the repo some time ago, but I don't remember why we deleted it. @themighty1 can you help here? any opinion against adding protobuffers to the repo?

Anyhow, if they need to be added, I'll add the ones that have been distributed with latest release.

Just out of curiosity O:) , may I ask you how are you using this repo as lib/package?

denysvitali:
> Just out of curiosity O:) , may I ask you how are you using this repo as lib/package?

I actually wanted to create a small web UI (I don't like the Python UI), but I later realized that the "UI Client" is in fact a client for the daemon, which in turn seems to be an hub for multiple "UI Clients" (computers?).

What I wanted to have is an UI for a single node. Do you have maybe an architecture overview of the project?

gustavo-iniguez-goya:
Nope. Some time ago I started a tui client, where I separated the daemon from the TUI logic with your idea in mind: allow to have multiple clients that connect to the server (GUI/TUI/web..). I commented out the code, and bear in mind that I wrote it in a few days and never reviewed it again .. O:)

https://github.com/gustavo-iniguez-goya/opensnitch/tree/cli/server/api
https://github.com/gustavo-iniguez-goya/opensnitch/tree/cli/server/cli

planetoryd:
why. they are generated

PR #625, opened 2022-02-19
Allow building on systems using musl

cinerea0:
Currently `queue.go` and `queue.h` use nonstandard type definitions which are unsupported by the musl libc implementation. This PR changes the definitions to be standard and allows the project to compile on musl.


gustavo-iniguez-goya:
thank you @cinerea0 :+1:

PR #684, opened 2022-07-01
ebpf: fix compiler issue

nberlee:
Signed-off-by: Nico Berlee <nico.berlee@on2it.net>


gustavo-iniguez-goya:
oh wow, sorry for that, thank you!

PR #686, opened 2022-07-02
position should be an uint64 in system-fw.json

nberlee:
fixes error in log: json: cannot unmarshal string into Go struct field FwRule.SystemRules.Rule.Position of type uint64

Signed-off-by: Nico Berlee <nico.berlee@on2it.net>

### DIFF
--- a/daemon/system-fw.json
+++ b/daemon/system-fw.json
@@ -7,7 +7,7 @@
         "Table": "mangle",
         "Chain": "OUTPUT",
         "Enabled": false,
-        "Position": "0",
+        "Position": 0,
         "Description": "Allow icmp",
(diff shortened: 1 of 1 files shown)


gustavo-iniguez-goya:
Hi @nberlee , the problem here is a little bit more tricky. The value is uint64 everywhere (daemon+proto+gui), but when the rules are saved on the GUI and sent back to the daemon, the value is sent as string.

As far as I can tell, the value is sent right as int from the daemon to the GUI, the GUI receives and stores it as int, but when sending back the rules after making any changes is when the problem occurs.

nberlee:
Yes you are right, needs more work, closing this for now as does not solve the cause.

PR #689, opened 2022-07-02
fw: support for icmpv6 nftables in system rules

nberlee:
- Add support for all available nftables ICMPv6 types (ip6tables -m icmpv6 --help)
- Build nftables ICMPv6 rules
- Create a default outbound ICMPv6 echo-request/reply rule
  (currently outbound echo-request ICMPv6 is by default denied)
```
$ doas nft list table inet mangle
table inet mangle {
	chain output {
		type filter hook output priority mangle; policy accept;
		icmp type { echo-reply, echo-request } accept
		icmpv6 type { echo-request, echo-reply } accept
		ct state related,new queue flags bypass to 0
	}
[...]
```


gustavo-iniguez-goya:
:heart:

PR #769, opened 2022-11-13
Here port should be also port in hungarian language.

szimszon:
Port translation was "kikötő" which means harbor and we need here "port" or "portszám" for port.


gustavo-iniguez-goya:
thank you!

szimszon:
uw

PR #780, opened 2022-12-21
daemon/Makefile: improvements to make distro packaging easier

craftyguy:
This is a handful of improvements to the daemon/Makefile that make it easier to package opensnitch on other distros. make's DESTDIR and PREFIX are widely used, and install is a very common way to handle installing/copying files/dirs to a location.

This *might* add some new dependency on `install` if some distro / environment doesn't include it, but it's a part of coreutils and busybox has a module for it... so it's likely everywhere already.


gustavo-iniguez-goya:
Thank you!

PR #782, opened 2022-12-22
docs: install instructions for dnf distros

mnahkies:
add command examples for installation / removal using `dnf` (eg: fedora)


gustavo-iniguez-goya:
thank you @mnahkies !

PR #826, opened 2023-01-31
Update workflow to go 1.19, actions bump, https chameleon, remove unused parameters

GiddyGoatGaming:
Not sure if it will act up on go 1.19 for workflow. It shouldn’t since it’s their “promise of compatibility” but you never know. 

The actions bumps won’t cause issues, other changes include upgrading chameleons link to https, and removing unused function parameters


gustavo-iniguez-goya:
Hi @GiddyGoatGaming !

Could you send the PRs grouped by modules? one PR for the actions, another one for the GUI, and for the legacy script.

Thank you.

PR #828, opened 2023-02-01
Bump workflow actions

GiddyGoatGaming:
Bumps actions to the latest in the workflow folder


GiddyGoatGaming:
> I'm using go 1.15 to test that the daemon builds with this version in particular, in order to build the packages for the releases. go 1.18 introduced a dependency (pthread_sigmask) that makes the binaries incompatible with older releases.
> 
> So I need automated tests with go 1.15.

Alright, sounds good. Il revert the go change and just request the workflow actions change

PR #829, opened 2023-02-01
Remove unused parameters

GiddyGoatGaming:
Removes unused parameters in to define paint cell


gustavo-iniguez-goya:
This change causes the GUI to crash. You need to update the call to paintCell() in order to match the function definition:

https://github.com/evilsocket/opensnitch/blob/e6edb18ee603f30bfc48af290c47d2c7b3a9178e/ui/opensnitch/actions/highlight.py#L140-L151

GiddyGoatGaming:
Ok, I’m not sure but is the change commit I just posted what you were asking?

gustavo-iniguez-goya:
Yep, that was it. Did you test the first commit? didn't it crash the gui on your system?

PR #862, opened 2023-02-26
Add New App Icon

KevDoy:
I'm proposing a new app icon due to the existing one being rather non-descript. However, I left the existing/tray icon included in this design to tie everything together. 

opensnitch-ui

Edit: Spelling


cypherbits:
Nice icon

staticssleever668:
Just a note: you can optimize these somewhat by using optipng, optionally removing some unneeded metadata with `--strip all`.

KevDoy:
updated per @staticssleever668's suggestion. Thanks.

KevDoy:
Resurfacing this
CC: @gustavo-iniguez-goya

Danny3:
I like the addition of a shield.
A door or doorman would also make sense to me for a firewall.
And I like the double arrows which also make sense to me for up / down connections.
Though it would make more sense to me if they were actually up and down and a lot of time we talk about upload and download, but I guess you wanted to still keep something from the current one.
I'm not very happy about seeing a cloud shape as I hate both the cloud and the AI.

And  I hate the flat design too, one color for a lot of things, no shadows, no lines, no grooves, no transitions.
I prefer more 3D-looking colored icons, something like Windows 7 had or the old Firefox icon had.

Anyway, thanks for trying to improve something about this program and for doing a pretty good job!

KevDoy:
Curious if there are any blockers to get this merged. 
Happy to make changes as well, just need to know.

Cheers

Danny3:
In the meantime, I would check how it looks when it's shown in the system tray as the smaller cloud + arrows would probably be too small to be legible.
And I think there's the need for another one, when some pop-ups have been missed and the default decision has been taken.
Right now there something in red appearing when that case happens.
And there's also one with II inside it in the system tray, when the program's protection is paused.

In case you were thinking that the program is using this new one for the system tray one too, besides the normal start menu and desktop shortcuts.

I wonder if there's any possibility in Linux add the icon as an option to choose from.
Maybe the program coming with both and they we choose from the right-click -> Properties menu.
In case the author doesn't want to change the default icon, that could be more acceptable.
And easier to do than having the author of the program make an option the in Preferences of the program to choose the icon.
If it's acceptable by you too.

ArjixWasTaken:
The shield is off center, that triggers my OCD.
I get that it's done to make space for the cloud, but the off-center ruins it for me.

PR #870, opened 2023-02-28
make version consistent with github tags

NRGLine4Sec:
Give the same version as the one obtained by the Github tags. In this way we can compare the result of the command `opensnitchd --version` with the github version (`curl --silent 'https://api.github.com/repos/evilsocket/opensnitch/releases' | grep 'tag_name.*rc.*'`).


gustavo-iniguez-goya:
Thank you @NRGLine4Sec :heart:

PR #873, opened 2023-03-01
Added Dutch translation

Vistaus:
🙂


gustavo-iniguez-goya:
Thank you @Vistaus !!

Could you leave out of the PR all the files except `ui/i18n/opensnitch_i18n.pro` and `ui/i18n/locales/nl_NL/opensnitch-nl_NL.ts` ?

The other translations and the files *.qm don't need to be updated.

By the way, if you use weblate most of the translators use it to translate the GUI: https://hosted.weblate.org/projects/opensnitch/gui-main-branch

Vistaus:
Oh, it automatically updated all those files when creating the Dutch file. But I've just uploaded the file to Weblate!

PR #889, opened 2023-03-26
Replace deprecated github.com/golang/protobuf

154pinkchairs:
I have also bumped Go version to 1.18, although many distros stopped providing updates to this version already and thus 1.19 would be even more preferred for a security program. I have also updated other dependencies and provided json tags for structs.


planetoryd:
@gustavo-iniguez-goya is there anything wrong with this PR.

PR #899, opened 2023-04-01
Wrap firewall errors in a common struct

154pinkchairs:
This allows a little bit of more flexibility since now there is only one return value instead of two. The underlying logic stays the same. 
Also unrelated, but I've noticed that you're not tracking go.sum. This isn't a good practice.

### DIFF
--- a/daemon/firewall/common/common.go
+++ b/daemon/firewall/common/common.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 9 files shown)


gustavo-iniguez-goya:
there're a couple of errors here:
https://github.com/evilsocket/opensnitch/blob/bd8e6ffae9b0cf168308f09a99382f72cc314fe4/daemon/firewall/iptables/rules.go#L23

https://github.com/evilsocket/opensnitch/blob/bd8e6ffae9b0cf168308f09a99382f72cc314fe4/daemon/firewall/iptables/rules.go#L32

err4 and err6 are not defined. Something in the line of `var err4, err6 error` would solve it.

Also "not enough return values" error here (func expects 2 values: nil, nil):
https://github.com/evilsocket/opensnitch/blob/bd8e6ffae9b0cf168308f09a99382f72cc314fe4/daemon/firewall/nftables/system.go#L138

and "common" package is not used error:
https://github.com/evilsocket/opensnitch/blob/bd8e6ffae9b0cf168308f09a99382f72cc314fe4/daemon/firewall/nftables/system.go#L6

154pinkchairs:
Seems like I've botched something when bisecting this with my other feature branch. It's my fault and I'll fix this. My apologies.

mjholub:
I'll take a look

mjholub:
Fixed. Ran the tests locally, they fail in the UI package, not sure why as they only print the cgo warninigs about deprecated dependencies

Issue #774, opened 2022-12-13
eBPF and compatibility with custom and hardened kernels (xanmod, liquorix, etc)


gustavo-iniguez-goya:
We've had several issues in the past regarding compatibility with this type of kernels and eBPF.

### Custom kernels (xanmod, others)

In order to use [eBPF as process monitor method](https://github.com/evilsocket/opensnitch/wiki/monitor-method-ebpf), the kernel must have some features enabled:

- [x] CONFIG_FTRACE=y
- [x] CONFIG_KPROBES=y  , CONFIG_KPROBES_ON_FTRACE=y, CONFIG_HAVE_KPROBES=y , CONFIG_HAVE_KPROBES_ON_FTRACE=y, CONFIG_KPROBE_EVENTS=y

Additionally:

- [x] CONFIG_HAVE_SYSCALL_TRACEPOINTS=y
- [x] CONFIG_FTRACE_SYSCALLS=y

You can check this options out with the following command: `$ grep FTRACE /boot/config-$(uname -r)` (or KPROBE, etc)

On the other hand, the path `/sys/kernel/debug/tracing/kprobe_events` must exist. Usually it's created when debugfs is mounted. If the path does not exist, one of the mentioned kernel features is not enabled.

### Hardened kernels / systems

There're some configurable kernel options that prevent us from working as expected (eBPF only, AFAIK). #767 
In particular these scripts seem to block some eBPF features: https://github.com/Kicksecure/security-misc#enhances-miscellaneous-security-settings

https://github.com/evilsocket/opensnitch/wiki/Known-problems#cannot-open-kprobe_events

---

**What happen if eBPF is not supported**

We fallback to `proc` monitor method. Starting from v1.6.0rc3 you'll receive an alert on the desktop. If you want to stop seeing those alerts: 1) you can change proc monitor method to `proc` (Preferences -> Nodes), or 2) run a kernel with needed requirements

Danny3:
Good that now an alert / notification will be shown to better explain what the problem is!

Maybe a link to this page or some page on the documentation with all these explanations could be added.

gustavo-iniguez-goya:
> liquorix kernels:
> ebpf proc monitor method works partially: connections are intercepted, but new/exit processes don't > (CONFIG_FTRACE_SYSCALLS not set).

[...]

PR #900, opened 2023-04-02
feat(daemon): add OpenRC script

154pinkchairs:
this PR adds a very basic script that would allow to get opensnitchd running under OpenRC init system. I've managed to get it to "work" i.e. not to crash, under Alpine 3.17 with Linux6.2.8-0-edge x86_64, but I keep getting this error:
```
Error creating queue #0: Error binding to queue: operation not permitted
```
I've set it to run using proc instead of bpf, because I have a hardened kernel. I can try giving it a test run in a VM without bpf hardening (I assume docker won't do in this case...). 
After this is merged and this issue resolved (I'll try to tackle this by myself, but any help is appreciated), I'd like to then proceed to package the application's components into an APKBUILD, so that it can be easily installed on Alpine.

### DIFF
--- a/daemon/opensnitchd-openrc
+++ b/daemon/opensnitchd-openrc
@@ -1,0 +1,36 @@
+#!/sbin/openrc-run
+# OpenSnitch firewall service
+
+depend() {
+    before net
+    after iptables ip6tables 
+    use logger
+    provide firewall
+}
+
+start_pre() {
+        /bin/mkdir -p /etc/opensnitchd/rules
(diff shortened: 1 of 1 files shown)


gustavo-iniguez-goya:
thank you @154pinkchairs ! sorry for the delay, I'm a bit offline lately.

gustavo-iniguez-goya:
> Error creating queue #0: Error binding to queue: operation not permitted

see if you have NFQUEUE support in your kernel, and in general all these options: https://github.com/evilsocket/opensnitch/issues/774

PR #959, opened 2023-06-06
Introduce 2 new daemon logging options: LogUTC & LogMicro.

lainedfles:
This request adds two new options to configure the daemon logging:

- **LogUTC**: toggles `time.Now().UTC()` vs `time.Now().local()`
- **LogMicro**: enable microsecond timestamp

Both options are exposed within the UI.

### DIFF
--- a/daemon/default-config.json
+++ b/daemon/default-config.json
@@ -9,6 +9,8 @@
     "InterceptUnknown": false,
     "ProcMonitorMethod": "ebpf",
     "LogLevel": 2,
+    "LogUTC": true,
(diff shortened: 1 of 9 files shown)


gustavo-iniguez-goya:
Thank you @lainedfles . For future changes: maybe it wouldn't be a bad idea to group logging options under "LogOptions": {} or something similar.

lainedfles:
That's a good suggestion. I'd also like to introduce a sub-tab under nodes dedicated to logging.

Issue #950, opened 2023-05-25
[Feature Request] Enable/disable autostart in tray


munix9:
Enable/disable autostart in tray.

### Summary:
I have created a patch to be able to enable/disable autostart in the tray.
It is a proof of concept (hack!) and some things can certainly be implemented better.

The following paths and files are defined as prerequisites for the checks and activation:
`/etc/xdg/autostart/opensnitch_ui.desktop` (optional)
`~/.config/autostart/opensnitch_ui.desktop`
`/usr/share/applications/opensnitch_ui.desktop`

This should all be standard, it works fine on openSUSE Tumbleweed (KDE Plasma, X11).

Maybe it can be used as a template to implement the functionality.

```
--- a/ui/opensnitch/service.py
+++ b/ui/opensnitch/service.py
@@ -181,11 +181,57 @@
         self._menu_enable_fw = self._menu.addAction(self.MENU_ENTRY_FW_DISABLE)
         self._menu_enable_fw.setEnabled(False)
         self._menu_enable_fw.triggered.connect(self._on_enable_interception_clicked)
+
+        self._menu.addSeparator()
+        self._menu_autostart = self._menu.addAction("Autostart")
+        self._menu_autostart.setCheckable(True)
+        self._menu_autostart.setChecked(self._is_autostart_enabled())
+        self._menu_autostart.triggered.connect(self._on_switch_autostart)
+        self._menu.addSeparator()
+
         self._menu.addAction(self.MENU_ENTRY_HELP).triggered.connect(
                 lambda: QtGui.QDesktopServices.openUrl(QtCore.QUrl(Config.HELP_CONFIG_URL))
                 )
         self._menu.addAction(self.MENU_ENTRY_CLOSE).triggered.connect(self._on_close)
 
+    def _is_autostart_enabled(self):
+        desktopFile = "opensnitch_ui.desktop"
+        systemAutostart = "/etc/xdg/autostart/" + desktopFile
+        if not os.path.isfile(systemAutostart) and os.path.isfile("/usr" + systemAutostart):
+            systemAutostart = "/usr" + systemAutostart
+        userAutostart = os.environ['HOME'] + "/.config/autostart/" + desktopFile
+        if os.path.isfile(userAutostart):
+            retVal = os.system("grep -qsi \"^Hidden=true\" " + userAutostart)
+            if retVal != 0:
+                return True
+        elif os.path.isfile(systemAutostart):
+            return True
+
+        return False
+
[...]

PR #964, opened 2023-06-09
xdg updates + autostart + lockfile

munix9:
I have summarized the various changes here:

- Use of pyxdg
- Extension of xdg.py (autostart)
- control autostart in tray
- check if os-ui is already running

See also https://github.com/evilsocket/opensnitch/issues/950

### DIFF
--- a/ui/bin/opensnitch-ui
+++ b/ui/bin/opensnitch-ui
@@ -40,11 +40,13 @@ if dist_path not in sys.path:
 from opensnitch.service import UIService
 from opensnitch.config import Config
 from opensnitch.utils import Themes, Utils, Versions
+from opensnitch.utils.xdg import xdg_runtime_dir
 import opensnitch.ui_pb2
 from opensnitch.ui_pb2_grpc import add_UIServicer_to_server
 
 def on_exit():
     server.stop(0)
+    lockfile.unlock()
     app.quit()
     sys.exit(0)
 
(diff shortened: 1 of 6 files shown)


gustavo-iniguez-goya:
This looks really good to me @munix9 , removing the pkill was a must.

Some general notes:

I see that pyxdg is packaged as python3-xdg on debian and ubuntu, and as python3-pyxdg/python36-pyxdg on fedora and opensuse.
Only available since ubuntu >= 20.04, so we would be incompatible with ubuntu 18 and 16 (and the linux mint/mx equivalents). Just a note, nothing really important, but worth mentioning.

ui/bin/opensnitch-ui
```diff
@@ -40,11 +40,13 @@ if dist_path not in sys.path:
 from opensnitch.service import UIService
 from opensnitch.config import Config
 from opensnitch.utils import Themes, Utils, Versions
+from opensnitch.utils.xdg import xdg_runtime_dir
 import opensnitch.ui_pb2
 from opensnitch.ui_pb2_grpc import add_UIServicer_to_server
 
 def on_exit():
     server.stop(0)
+    lockfile.unlock()
```
gustavo-iniguez-goya, line 49: on my system this complains about `lockfile is not defined`
a missing import?

munix9:
> This looks really good to me @munix9 , removing the pkill was a must.
> 
> Some general notes:
> 
> I see that pyxdg is packaged as python3-xdg on debian and ubuntu, and as python3-pyxdg/python36-pyxdg on fedora and opensuse. Only available since ubuntu >= 20.04, so we would be incompatible with ubuntu 18 and 16 (and the linux mint/mx equivalents). Just a note, nothing really important, but worth mentioning.

This is not nice. Maybe I can integrate the needed methods of pyxdg directly into xdg.py to avoid the dependency.
I'll have a look at it.

ui/bin/opensnitch-ui
munix9: That's strange, it works for me without any problem.

`lockfile` is defined in line 99 https://github.com/evilsocket/opensnitch/blob/d003cc162c56db1d69bd62265ef3b5216dac4a05/ui/bin/opensnitch-ui#L99, directly after `app`.

What versions (python, qt) are you using?

gustavo-iniguez-goya: nah, nevermind, my local branch of the PR was not updated, sorry.

munix9:
@gustavo-iniguez-goya pyxdg dependency is removed.

gustavo-iniguez-goya:
@munix9 , did you see this error? it occurs when the `~/.config/autostart/opensnitch_ui.desktop` already exists. We'll hit this error when users coming from v1.5.x install the next version.

```python
  File "/usr/lib/python3/dist-packages/opensnitch/service.py", line 202, in _on_switch_autostart
    self._autostart.enable(self._menu_autostart.isChecked())
  File "/usr/lib/python3/dist-packages/opensnitch/utils/xdg.py", line 82, in enable
    shutil.copyfile(self.systemAutostart, self.userAutostart)
  File "/usr/lib/python3.8/shutil.py", line 244, in copyfile
    raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
shutil.SameFileError: '/etc/xdg/autostart/opensnitch_ui.desktop' and '/home/ga/.config/autostart/opensnitch_ui.desktop' are the same file
```

gustavo-iniguez-goya:
thank you @munix9 !

PR #968, opened 2023-06-12
use `uname -m` instead of the `arch` command

nnsee:
The `arch` command is deprecated on modern systems, and indeed, many distros do not provide it (one of those being Arch Linux, ironically). Since `uname` is already used in the Makefile, prefer its `-m` flag equivalent instead.


nnsee:
For context - I am the maintainer of the [opensnitch-ebpf-module](https://aur.archlinux.org/packages/opensnitch-ebpf-module) AUR package. Currently, I'm passing `ARCH` manually, as can be seen in the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=opensnitch-ebpf-module#n23), but it would be great if I didn't have to do this.

PR #969, opened 2023-06-17
Dinit service file

jackffmm:
Dinit service file


gustavo-iniguez-goya:
thank you @jackffmm !

jackffmm:
Thank  you!
just a note, i changed the restart delay from 30s to 15s. Why is it so long on the systemd service?

gustavo-iniguez-goya:
I think because it's the default value in many examples. Right now the if the daemon has not been stopped after 10s, it exits without waiting for other tasks to finish.

Issue #940, opened 2023-05-16
[Feature Request] Allow to start UI in background even, when tray is not available.


WojtekWidomski:
Allow to start UI in background (without showing statistics window) even, when tray is not available, for example using command line option or option in settings.

### Summary:
It is not possible to automatically start opensnitch-ui on login in background without installing gnome-shell-extension-appindicator and displaying tray icon (I do not like to have too many icons on panel and when I want to open statistics window I can do it from applications menu.)

gustavo-iniguez-goya:
added

PR #975, opened 2023-06-24
Allow starting in background when tray not available

WojtekWidomski:
This PR adds `--background` command line argument that starts GUI in background even when tray not available and uses QLocalSocket and QLocalServer instead of lockfile (now it is possible to open window without starting new instance, using launcher in programs menu).

Please review this code, because I never used Qt.

This will fix issue #940.


gustavo-iniguez-goya:
hey @WojtekWidomski , thank you for looking into this.

What is the purpose of the change, launch the GUI when the systray is not available?

WojtekWidomski:
@gustavo-iniguez-goya I don't want tray icon, but when OpenSnitch is added to startup applications and gnome-shell-extension-appindicator is disabled, OpenSnitch window shows on every login.
After this change it is possible to add `opensnitch-ui --background` to startup applications.

gustavo-iniguez-goya:
Thank you @WojtekWidomski !

I'll rework it a little bit, but it improves what @munix9 solved with the lockfile and at the same time solved the problem of displying the GUI on demand when the systray is not available. Thank you both!

PR #985, opened 2023-07-07
Use temporary files instead of piping in ebpf Makefile

nnsee:
Currently, when compiling the eBPF module, in the Makefile, the output of `clang` is directly piped to `llc`. This is bad, because in this case, should `clang` fail, `llc` still runs, but produces faulty `.o` files. `make` never picks up on it and doesn't abort compilation - a success is reported across the board (apart from the `clang` log).

This patch removes the file and explicitly writes (and later deletes) `clang`'s output to a `.partial` file. This is later used as the input of `llc`.


gustavo-iniguez-goya:
thank you @nnsee :) everyone seems to use pipes, but it's true that empty/invalid objects was a problem.

PR #1011, opened 2023-08-09
Introduce Sqlite WAL journal mode

lainedfles:
- Add [Write-Ahead Logging](https://www.sqlite.org/wal.html) option for file mode database
- Disable by default
- Migrate journal_mode and related `PRAGMA` statement handling from _create_tables() to new functions: get_journal_mode()/set_journal_mode()

This yields a significant reduction of constant IO writes on my systems when db file mode is enabled. 

Note that `PRAGMA synchronous = NORMAL` doesn't appear to activate properly. This problem occurs on master branch builds as well. Manual inspection of the db shows that synchronous always = FULL unless manually set directly but this warrants additional discussion elsewhere.


gustavo-iniguez-goya:
Thank you @lainedfles !

I experienced a lot of DB corruptions using WAL and other modes when I added the option to save events to disk.
That's why I ended up using previous options.

But things has changed a lot since then, and allowing to configure it is great :)

Issue #1018, opened 2023-08-15
[Feature Request] [Enhancement] Systemd service file


TriMoon:
### Summary:
The systemd service unit can be enhanced to start **BEFORE** any network is configured, and thus allow interception and protection at an earlier stage...

### Service file:
Below is my crafted service file:
```ini
# /etc/systemd/system/opensnitchd.service
[Unit]
Description=Application firewall OpenSnitch
Documentation=https://github.com/gustavo-iniguez-goya/opensnitch/wiki
Documentation=man:systemd.special
Documentation=man:systemd.service
Documentation=man:systemd.exec
Documentation=man:systemd.unit

DefaultDependencies=no
Before=network-pre.target shutdown.target
Wants=network-pre.target
Conflicts=shutdown.target
# Don't start when 'no-appfw` is in kernel command-line, to allow booting without it.
ConditionKernelCommandLine=!no-appfw

[Service]
Type=exec
ConfigurationDirectory=%N/rules
ConfigurationDirectoryMode=0700

Environment='custom_cfg=%E/%N/rules'
# Environment='opts=-debug'

ExecCondition=%N -check-requirements
ExecStart=%N -rules-path $custom_cfg $opts

# Signal-info was taken from the init.d script, but it just exits and then systemd restarts the service...
ExecReload=kill -HUP $MAINPID
Restart=always
RestartSec=30
TimeoutStopSec=10
# Ensure it is not killed by the Linux kernel's Out-Of-Memory (OOM) killer.
# https://www.freedesktop.org/software/systemd/man/systemd.exec.html#OOMScoreAdjust=
OOMScoreAdjust=-1000

[Install]
WantedBy=basic.target
```

---
Notable changes/additions:

1. **Unit filename:**
:warning: The service filename was changed to `opensnitchd.service` !!!
This was needed for automatically using the unit name in the service definition using the `%N` specifier...
See: [Specifiers](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Specifiers)@man `systemd.unit#Specifiers`
   - _The previous filename was lacking the `d` in it's name !_
   - Thus if you want to use this service file while **at same time** having the _currently distributed_ unit file in your system you need to "mask" the old one to prevent it from being used and started:
     ```bash
     sudo systemctl mask --now opensnitch.service
     ```

1.  **Unit ordering:**
Made sure the daemon starts before any network related devices or services are created/started by using `network-pre.target`, See:
[...]

PR #1019, opened 2023-08-15
Update opensnitchd.service

TriMoon:
See: #1018

### DIFF
--- a/daemon/opensnitchd.service
+++ b/daemon/opensnitchd.service
@@ -1,15 +1,37 @@
 [Unit]
 Description=Application firewall OpenSnitch
-Documentation=https://github.com/evilsocket/opensnitch/wiki
+Documentation=https://github.com/gustavo-iniguez-goya/opensnitch/wiki
+Documentation=man:systemd.special
+Documentation=man:systemd.service
+Documentation=man:systemd.exec
+Documentation=man:systemd.unit
+
+DefaultDependencies=no
+Before=network-pre.target shutdown.target
+Wants=network-pre.target
+Conflicts=shutdown.target
+# Don't start when 'no-appfw` is in kernel command-line, to allow booting without it.
+ConditionKernelCommandLine=!no-appfw
 
 [Service]
-Type=simple
-PermissionsStartOnly=true
-ExecStartPre=/bin/mkdir -p /etc/opensnitchd/rules
-ExecStart=/usr/local/bin/opensnitchd -rules-path /etc/opensnitchd/rules
+Type=exec
+ConfigurationDirectory=%N/rules
+ConfigurationDirectoryMode=0700
+
+Environment='custom_cfg=%E/%N/rules'
+# Environment='opts=-debug'
+
+ExecCondition=%N -check-requirements
+ExecStart=%N -rules-path $custom_cfg $opts
+
+# Signal-info was taken from the init.d script, but it just exits and then systemd restarts the service...
+ExecReload=kill -HUP $MAINPID
 Restart=always
 RestartSec=30
 TimeoutStopSec=10
+# Ensure it is not killed by the Linux kernel's Out-Of-Memory (OOM) killer.
+# https://www.freedesktop.org/software/systemd/man/systemd.exec.html#OOMScoreAdjust=
+OOMScoreAdjust=-1000
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=basic.target


daemon/opensnitchd.service
```diff
@@ -1,15 +1,37 @@
 [Unit]
 Description=Application firewall OpenSnitch
-Documentation=https://github.com/evilsocket/opensnitch/wiki
+Documentation=https://github.com/gustavo-iniguez-goya/opensnitch/wiki
```
TriMoon, line 3: Hmm this line needs tobe undone it seems :rofl:

TriMoon: fixed now :smiley_cat:

TriMoon:
:warning:
The service files as [distributed in the `deb` package](https://github.com/evilsocket/opensnitch/releases/tag/v1.6.2) still has **WRONG name** !!!
Someone needs to fix the packaging script !

TriMoon:
I think i found the spot where the packaged version in the `deb` differs from what is actually meant to be used:
- https://github.com/evilsocket/opensnitch/blob/master/utils/packaging/daemon/deb/debian/opensnitch.service

@gustavo-iniguez-goya could you please make that file a symbolic link to the one in the master tree at: https://github.com/evilsocket/opensnitch/blob/master/daemon/opensnitchd.service :thinking: 
It will eliminate further discrepancies between the packaged version and the sources of the daemon...

daemon/opensnitchd.service
```diff
@@ -1,15 +1,37 @@
...
-Type=simple
-PermissionsStartOnly=true
-ExecStartPre=/bin/mkdir -p /etc/opensnitchd/rules
-ExecStart=/usr/local/bin/opensnitchd -rules-path /etc/opensnitchd/rules
+Type=exec
+ConfigurationDirectory=%N/rules
+ConfigurationDirectoryMode=0700
+
+Environment='custom_cfg=%E/%N/rules'
+# Environment='opts=-debug'
```
TriMoon, line 22: > @lainedfles There is a benign warning if `$opts` is unset. I'd recommend setting an empty variable as a better practice. Like: `Environment='opts='`

Nice catch, i expected for it to become empty in the expansion so i didn't expect that :+1:

TriMoon:
> > The introduction of the _custom_config_ variable may be unnecessary unless use of `EnvironmentFile=` is planned. The structure of systemd makes it trivial to override `ConfigurationDirectory=` which, since v240, will expose the `$CONFIGURATION_DIRECTORY` environment variable automatically that we can re-use.
> 
> It can be removed when the functionality you mention is implemented, but until then lets keep it in the unit.


> +ExecStart=/usr/bin/%N -rules-path $CONFIGURATION_DIRECTORY $opts

Aha THAT's how you planned on using it?
Sure we could change it that way if you want... :woman_shrugging: 
- https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory= :+1:

daemon/opensnitchd.service
```diff
@@ -1,15 +1,37 @@
...
+
+ExecCondition=%N -check-requirements
+ExecStart=%N -rules-path $custom_cfg $opts
```
TriMoon, line 25: > > The introduction of the _custom_config_ variable may be unnecessary unless use of `EnvironmentFile=` is planned. The structure of systemd makes it trivial to override `ConfigurationDirectory=` which, since v240, will expose the `$CONFIGURATION_DIRECTORY` environment variable automatically that we can re-use.
> 
> It can be removed when the functionality you mention is implemented, but until then lets keep it in the unit.


> +ExecStart=/usr/bin/%N -rules-path $CONFIGURATION_DIRECTORY $opts

Aha THAT's how you planned on using it?
Sure we could change it that way if you want... :woman_shrugging: 
- https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory= :+1:

daemon/opensnitchd.service
TriMoon, line 22: > @lainedfles There is a benign warning if `$opts` is unset. I'd recommend setting an empty variable as a better practice. Like: `Environment='opts='`

Nice catch, i expected for it to become empty in the expansion so i didn't expect that :+1:

daemon/opensnitchd.service
TriMoon, line 25: > @lainedfles 
> > The introduction of the _custom_config_ variable may be unnecessary unless use of `EnvironmentFile=` is planned. The structure of systemd makes it trivial to override `ConfigurationDirectory=` which, since v240, will expose the `$CONFIGURATION_DIRECTORY` environment variable automatically that we can re-use.
> 
> It can be removed when the functionality you mention is implemented, but until then lets keep it in the unit.


> +ExecStart=/usr/bin/%N -rules-path $CONFIGURATION_DIRECTORY $opts

Aha THAT's how you planned on using it?
Sure we could change it that way if you want... :woman_shrugging: 
- https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory= :+1:

TriMoon (commented):
Sugestions from @lainedfles

TriMoon (commented):
Keep in mind that replacing `custom_cfg` with `$CONFIGURATION_DIRECTORY` this way will prevent the admin from overriding, the rules directory used, using drop-in's if (s)he chooses so.

If people like the idea of being able to override like that, we could undo **THAT** change...

TriMoon:
> > `WantedBy=basic.target` may become problematic for [initrd boot](https://www.freedesktop.org/software/systemd/man/bootup.html#Bootup%20in%20the%20initrd) because basic.target is reached before any filesystems are mounted. However, I think that the directives under `[Unit]` should catch this case under most circumstances so it may be a non-issue but worth mention.
> 
> The `opensnitchd` service/daemon is not started inside the initrd so that's not an issue. Also the dependencies used under `[Unit]` make sure it is started way after `/etc` is mounted anyhow.

If things do go wrong or just to be extra specific we could add `After=local-fs.target`, but i don't forsee any need at moment. :wink:

luzpaz:
Any progress here ?

TriMoon:
If i would have made any more changes they would be shown in this thread as commits so no, the PR is still waiting for merging/acceptance...

daemon/opensnitchd.service
```diff
@@ -1,15 +1,37 @@
...
+Documentation=man:systemd.service
+Documentation=man:systemd.exec
+Documentation=man:systemd.unit
+
+DefaultDependencies=no
+Before=network-pre.target shutdown.target
+Wants=network-pre.target
+Conflicts=shutdown.target
+# Don't start when 'no-appfw` is in kernel command-line, to allow booting without it.
+ConditionKernelCommandLine=!no-appfw
```
petterreinholdtsen, line 14: Is  no-appfw a well known kernel command line option, or introduced here?

petterreinholdtsen:
Note, Debian uses fragments from this patch in [1030-systemd-service-earlier.patch](https://sources.debian.org/src/opensnitch/1.6.9-2/debian/patches/1030-systemd-service-earlier.patch).  It would be nice to know if at least these parts will make it into master.

LordGrimmauld:
In NixOS we currently hand-roll our own minimal version of the systemd service here, without any of these patches: https://github.com/NixOS/nixpkgs/blob/ee930f9755f58096ac6e8ca94a1887e0534e2d81/nixos/modules/services/security/opensnitch.nix#L202-L245

I was intending to add the systemd hardening options to opensnitch, and i was considering upstreaming those - that would result in a merge conflict on this file, but is mostly an orthogonal change set in terms of logic.

daemon/opensnitchd.service
TriMoon: I invented it just for this service :wink:

PR #1024, opened 2023-08-19
Ubuntu

ozialien:
Small script to compile the whole thing in a container until getting around to an apt/rpm installer.


ozialien:
Bad files made it in.

PR #1130, opened 2024-05-12
Fix SIGSEGV caused by uninitialized firewall and logger

1fishe2fishe:
I compile opensnitch from git, and I got a journalctl backlog full of these today:
image
This seemed to be related to bc3209494524497566a72a973f5b9ce1df36e67f, and after restoring firewall.Init(), I got this:
image
[This change](https://github.com/evilsocket/opensnitch/commit/0b67c1a429b7c155da50ef7056737c521cfb3c91#diff-b8e796a8bd84a7996ec85b5d42c6766c6e4cb40ef3f21869d914e66ed74f2d39R110) to daemon/ui/config_utils.go, where SetLoggers() is called, changes it to only call it on one branch, so that seemed to be where the null logger problem was introduced.

My change seems to work properly without issues, but I'm not a Go person so feel free to correct me.

### DIFF
--- a/daemon/main.go
+++ b/daemon/main.go
@@ -596,7 +596,14 @@ func main() {
 
 	setupWorkers()
 	setupQueues()
-
+	if err = firewall.Init(
+		uiClient.GetFirewallType(),
+		cfg.FwOptions.ConfigPath,
+		cfg.FwOptions.MonitorInterval,
+		&queueNum); err != nil {
+		log.Warning("%s", err)
+		uiClient.SendWarningAlert(err)
+	}
 	// queue and firewall rules should be ready by now
 
 	uiClient.Connect()
--- a/daemon/ui/config_utils.go
+++ b/daemon/ui/config_utils.go
@@ -107,10 +107,10 @@ func (c *Client) reloadConfiguration(reload bool, newConfig config.Config) *moni
 		log.Debug("[config] reloading config.server.loggers")
 		c.loggers.Stop()
 		c.loggers.Load(newConfig.Server.Loggers)
-		c.stats.SetLoggers(c.loggers)
 	} else {
 		log.Debug("[config] config.server.loggers not changed")
 	}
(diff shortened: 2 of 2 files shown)


gustavo-iniguez-goya:
hey @1fishe2fishe , thank you for taking a look at this!

Could you verify that latest sources compile fine? I added more changes after this commit, sorry for that O:)

1fishe2fishe:
Well, it does compile fine with the latest commit, but the problem still lingers. I think the fix is still needed, and my patch doesn't have conflicts with the master branch (at least as of c9ad9005e346b5e2bcf21311b2db8dec47985890), so I guess it's good to merge.
image

1fishe2fishe:
That is to say, with my patch it runs just fine and it successfully works and connects to the UI without segfaulting or anything.
image

gustavo-iniguez-goya:
no problem with the patch, it actually fixes the crashes, but let's fix it more in line with the latest changes (avoiding config reloading as much as possible).

Could you try the following changes and let me know ?

add `stats.SetLoggers(loggerMgr)`
https://github.com/evilsocket/opensnitch/blob/c9ad9005e346b5e2bcf21311b2db8dec47985890/daemon/main.go#L593-L595

and remove "reload" here:
https://github.com/evilsocket/opensnitch/blob/c9ad9005e346b5e2bcf21311b2db8dec47985890/daemon/ui/config_utils.go#L168

1fishe2fishe:
That compiled and ran without issues, thanks for fixing the problem.
I don't have a good understanding of this codebase/language, so I'll close the PR and issue in the meantime and wait for you to push the more efficient fix instead of what is essentially a commit revert.

gustavo-iniguez-goya:
Thank you @1fishe2fishe for taking a look at this and report it! These changes have been waiting for a long time, and there will be more, so more testing and unit tests will be needed.

PR #1135, opened 2024-05-18
Improve wording in Top menu

ponychicken:
I assume that the main gui window was just displaying statistics.
Now you can configure rules, and do other various settings, the menu point "Statistics" doesn't seem appropriate anymore..


gustavo-iniguez-goya:
exactly @ponychicken :) thank you!

PR #1179, opened 2024-08-29
remove lagacy cname_trackers.txt reference from update_adlists.sh

abuturabofficial:
AdGuard no longer maintains/supports that file type, and the link points to a 404 error page. After removing all the references, the user will still be able to update their local blocklists.

### DIFF
--- a/utils/scripts/ads/update_adlists.sh
+++ b/utils/scripts/ads/update_adlists.sh
@@ -36,24 +36,6 @@ adsListNames=(
     "adaway-hosts.txt"
     "yoyo-adservers.txt")
 
-function download_cname_trackers
-{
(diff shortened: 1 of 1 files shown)


abuturabofficial:
The legacy link: https://raw.githubusercontent.com/AdguardTeam/AdguardFilters/master/SpywareFilter/sections/cname_trackers.txt

AdGuard now splits the filters into multiple files, there is currently no list that merges them together available by Adguard.
https://github.com/AdguardTeam/AdguardFilters/tree/master/SpywareFilter/sections

gustavo-iniguez-goya:
thank you @abuturabofficial ! I'll take this into account in case this script is modified.

PR #1185, opened 2024-09-09
ebpf_prog/Makefile: temp files are now handled by a pattern rule

ariel-anieli:
* `%.bc` are autoremoved: [these LLVM IR files are intermediate](https://www.gnu.org/software/make/manual/html_node/Chained-Rules.html)
* `%.o` are now produced by a wildcard search
* introduced `.SUFFIXES:` [for cleaning up the implicit rules](https://www.gnu.org/software/make/manual/html_node/Suffix-Rules.html)
* else Makefile would have generated `%.o` from its own database.


gustavo-iniguez-goya:
Thank you @ariel-anieli !

PR #1195, opened 2024-09-23
Add Indonesian translation

atriwidada:
- add Indonesian translation
- sort translation files in opensnitch_i18n.pro


gustavo-iniguez-goya:
thank you @atriwidada :heart:

PR #1209, opened 2024-10-17
Make example explanation consistent with regex - take #2

atriwidada:


gustavo-iniguez-goya:
thank you @atriwidada !

PR #1235, opened 2024-12-10
fix #1234: reduce table row vertical size

### LINKED ISSUE #1234: [Feature Request] Reduce vertical row size for data tables
### Summary:
There is excessive padding for lines of text in the data tables (Events, Rules, etc), half the screen is empty space, not ideal for viewing data. Add UI setting for reducing vertical padding between lines of text

e3dio:
fix #1234

# Before:

opensnitch_spacing

# After:

os_space


gustavo-iniguez-goya:
Setting a fixed height for the rows breaks the look&feel in some configurations https://github.com/evilsocket/opensnitch/issues/1234#issuecomment-2531465398

PR #1237, opened 2024-12-11
Lan access control

nolancarougepro:
In the daemon and ui folders, there is a file named network_aliases.json.

image

For the ui folder, there is a file network_aliases.py that handles interactions with the JSON file, such as retrieving a list of IPs associated with an alias. When we check whether an IP is present in the JSON file, if it is found, we add a line in the prompt.

image

When we click on LAN, the rules are saved as follows:

image

If we want to modify the rules, we print each possible alias:

image

The rules are then saved as follows:

image

For the daemon part: The JSON file is loaded when the daemon starts. For each connection, we check whether the rule is of type "network". If there is an alias and the destination IP is in the list from the JSON file, we use this information to proceed with the desired action.

I decided to keep this code for backward compatibility, but it can easily be deleted:
```python
    classA_net = r'10\.\d{1,3}\.\d{1,3}\.\d{1,3}'
    classB_net = r'172\.1[6-9]\.\d+\.\d+|172\.2[0-9]\.\d+\.\d+|172\.3[0-1]+\.\d{1,3}\.\d{1,3}'
    classC_net = r'192\.168\.\d{1,3}\.\d{1,3}'
    others_net = r'127\.\d{1,3}\.\d{1,3}\.\d{1,3}|169\.254\.\d{1,3}\.\d{1,3}'
    multinets = r'2[32][23459]\.\d{1,3}\.\d{1,3}\.\d{1,3}'
    MULTICAST_RANGE = "^(" + multinets + ")$"
    LAN_RANGES = "^(" + others_net + "|" + classC_net + "|" + classB_net + "|" + classA_net + "|::1|f[cde].*::.*)$"
    LAN_LABEL = "LAN"
    MULTICAST_LABEL = "MULTICAST"
```


gustavo-iniguez-goya:
thank you @nolancarougepro !

I've added a few comments.

Idea for future improvements:

image


Add the LAN item to the DstIP combobox:



That way you can filter for example by executable + LAN

nolancarougepro:
Thank for the feedback.
I will work on it.

nolancarougepro:
I updated the code as mentioned. It now looks like this when an application tries to reach an alias (in this case, the LAN alias):
image
image

PR #1282, opened 2025-02-28
Update bug_report.md

NormPlum:
The Bug Report template was missing the 'bug' label, and there's quite a few bits of help text that can be hidden (to prevent users needing to do it themselves). Made these changes to bring it more in line with the current Feature Request template.


gustavo-iniguez-goya:
thanks! :heart:

PR #1309, opened 2025-04-06
Wiki: More clear explanation on rule precedence

mittei:
I found the Wiki to be a little confusing about how rules are selected and I had to check the code to understand it. This PR makes the Wiki more understandable (hopefully).

Relevant code:
https://github.com/evilsocket/opensnitch/blob/297ffbbb420b7b3719eb1604194be25d79c3b6b4/daemon/rule/loader.go#L485-L504
https://github.com/evilsocket/opensnitch/blob/297ffbbb420b7b3719eb1604194be25d79c3b6b4/daemon/main.go#L444-L446

### DIFF
--- a/wiki/Rules.md
+++ b/wiki/Rules.md
@@ -80,19 +80,19 @@ Rules are stored as JSON files inside the `-rule-path` folder, in the simplest c
 - If you select multiple lists on the same rule, bear in mind that the connections you want to match must
  [Read this disccussion to learn more](https://github.com/evilsocket/opensnitch/discussions/877#discussioncomment-5247997)
 
-- By default Deny rules take precedence over the rest of the rules. If a connection match a Deny rule, opensnitch won't continue evaluating rules.
+- Rule precedence: When a connection is attempted, OpenSnitch evaluates each of the defined rules in the rule name's alphabetical order (since v.1.2.0). As soon as it encounters a  Deny/Reject rule or an _Important_ ([x] Priority) rule (since v1.2.0) that matches the connection, that rule will be immediately selected as the effective rule. If no such rule is found, then the last non-Important Allow rule that matched will be selected. If no rule matched, it shows a pop-up dialogue, or applys the default action if that's not possible.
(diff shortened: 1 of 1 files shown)


mittei:
One thing I still don't understand after looking at the code is this: Even though I set the default action to "allow", when a new connection (that doesn't match any rule) is made while a pop-up is already shown, the new connection doesn't get allowed for some reason.

image

But looking at this code, it looks like it should apply the default action when a pop up is already open.
https://github.com/evilsocket/opensnitch/blob/297ffbbb420b7b3719eb1604194be25d79c3b6b4/daemon/main.go#L451-L457

I'd appreciate it if someone explains this.

mittei:
Thanks for merging!

If anyone could answer my question above, I'd appreciate it

gustavo-iniguez-goya:
sorry @mittei , I've been away a few weeks.

> Not super related to this PR but, one thing I still don't understand after looking at the code is: Even though I set the default action to "allow", when a new connection (that doesn't match any rule) is made while a pop-up is already shown, the new connection doesn't get allowed for some reason.

When the daemon connects to the GUI, the GUI has a DefaultAction which overwrites daemon's DefaultAction:

image

service.py -> Subscribe -> overwrite_nodes_config

https://github.com/evilsocket/opensnitch/blob/624d62864119b02fc63a1c8ab9fae51d5e0b5114/ui/opensnitch/service.py#L702-L705

Users were confused, because the DefaultAction was set to Deny in the GUI, but requests were being allowed while a popup was active.

On the other hand, there's the daemon's DefaultAction (Preferences -> Nodes), which will take effect when it's not connected to the GUI.

mittei:
@gustavo-iniguez-goya Thanks for the reply, but I just did the following and didn't get the expected result:

1. Set the Pop-ups Default Action to Allow (screenshot below). Based on what you said, this should overwrite the demon's default action to allow
2. Run `curl <url1>` -> Pop up appears
3. Run `curl <url2>` while the pop up is still open
4. `curl <url2>` doesn't get any response for some reason

image

(It's not like I actually want allow-by-default in my setup but I just want to understand)

PR #1320, opened 2025-04-16
ui,stats: handle dict-wrapped notifs in callback

lainedfles:
Modify the `_cb_notification_callback` method to properly process notifications that are wrapped in a dictionary to ensure consistent handling regardless of how they're sent from different parts of the codebase.


lainedfles:
Resolves #1321

gustavo-iniguez-goya:
Thank you @lainedfles !

I've reproduced it with the AUR package, but for some reason on Ubuntu 22 works fine.

lainedfles:
Thanks @gustavo-iniguez-goya, I was also able to reproduce it with the manual installation instructions. I wonder if it's something that Ubuntu/Debian is fixing with a patch or something that Arch Linux has broken. I wasn't able to isolate the cause but I can tell you that Ubuntu 24.04 with updated dependencies doesn't crash either.

PR #1324, opened 2025-04-18
Introduce new 12 hour duration option.

### LINKED ISSUE #1286: [Feature Request] Longer duration for temporary rules - 8 hours
### Summary:

Sometimes I have real work to do and don't want to bother *today* figuring out what domains to allow and which to deny, so I keep clicking "1h" in the drop-down, and keep getting the pop-ups.

An "8h" setting would solve this. No need for "1 day" - too much, and it would clutter the drop-down.

Image

petterreinholdtsen:
Fixes #1286

### DIFF
--- a/ui/opensnitch/config.py
+++ b/ui/opensnitch/config.py
@@ -66,6 +66,7 @@ class Config:
     DURATION_UNTIL_RESTART = "until restart"
     DURATION_ALWAYS = "always"
     DURATION_ONCE = "once"
+    DURATION_12h = "12h"
     DURATION_1h = "1h"
     DURATION_30m = "30m"
     DURATION_15m = "15m"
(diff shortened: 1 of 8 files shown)


petterreinholdtsen:
Is this the correct approach for adding a new duration option?  Is it a problem that the '6' value change behaviour from until reboot to '12 hours', or will it cause problem with existing setups?  The patch is just a untested draft so far.

gustavo-iniguez-goya:
tested and works as expected. 

If users had customized the Defaultduration in the GUI to "until reboot", it'll default now to 12h for them.

Ideally we should have a predefined list of options, and let the users add more options dynamically. Otherwise users will keep asking for more options (24h, 1day, 7days, 15days, 7.54 days, etc..).

PR #1327, opened 2025-04-20
Changed how ebpf build find kernel headers from running to installed version.

petterreinholdtsen:
This get the build working when building in a chroot or when the running kernel is different from the kernel headers installed.


gustavo-iniguez-goya:
thank you @petterreinholdtsen ! :heart:

PR #1332, opened 2025-04-23
refactor: Update localhost rules and add IPv6

lainedfles:
- Changed the local host rule from using regexp to network operator
- Updated destination IP check from specific addresses to CIDR notation (127.0.0.0/8)
- Added new rule for IPv6 localhost connections (::1/128)
- Added nolog property to prevent logging of localhost traffic


gustavo-iniguez-goya:
thank you @lainedfles , I didn't even remember adding these rules to the repo.

PR #1344, opened 2025-05-18
ui: remove unused imports

ddogfoodd:
> [!WARNING]  
> I didn't actually test the changes, so a test/review would be very much appreciated and required before a potential merge.

My LSP showed some unused imports in the ui source.
For each of the files changed, I searched the corresponding file for the appearance of the seemingly unused import. If it didn't occur anywhere except in the import statement, I removed it.

There were some more occurences where imports currently seem unused but their usage is just commented out at the moment, I didn't remove such.


gustavo-iniguez-goya:
tested and works fine, thank you @ddogfoodd

PR #1359, opened 2025-06-05
check whether dist_path exists

munix9:
"dist_path" seems to be a debian thing, in openSUSE for example the path does not exist.

### DIFF
--- a/ui/bin/opensnitch-ui
+++ b/ui/bin/opensnitch-ui
@@ -35,7 +35,7 @@ from concurrent import futures
 import grpc
 
 dist_path = '/usr/lib/python3/dist-packages/'
-if dist_path not in sys.path:
+if os.path.exists(dist_path) and dist_path not in sys.path:
     sys.path.append(dist_path)
 
 from opensnitch.service import UIService


gustavo-iniguez-goya:
Thanks @munix9 .

Yes, that path doesn't exist on other distros. The real problem is packaging the app for different distros regardless the python version.

For example, if you're on OpenSuse with python3.12, and the system upgrades to python3.13, python3 will point to python3.13 which will try to find opensnitch under `/usr/lib/python3.13/`.

If opensnitch is packaged by the distro, this should not be a problem. But in the meantime, we have to ensure that upgrading to newer python versions doesn't break the GUI.

We had this issue in the past, when transitioning from python3.9 to python3.10 or a similar scenario.

munix9:
Yes, you're right.

For this reason, when openSUSE creates the rpm packages, care is taken to ensure that the python version is explicitly requested during installation that was also used when building the package (should also be the case with Fedora, for example).

E.g. `/usr/bin/opensnitch-ui` is then adapted accordingly:
from `#!/usr/bin/env python3` to `#!/usr/bin/python3.13`

Of course, this is difficult if you offer rpm packages for any distribution, as each distribution may have a slightly different approach.

PR #1396, opened 2025-07-16
Add Operand process.parent.command

b2ag:
Hi,

I want to distinguish between different Electron apps that share the same Electron binary path by matching on their parent processes' command line. When started like `electron app.asar`, Electron seems to fork some child processes. Those processes are missing the `app.asar` bit of the command line and are therefor hard to map to the root application.

I'm aware that command lines can be manipulated but can't think of a better way for doing this.

I'm also not sure if `ReadCmdline` should be called from `procmon/details.go: GetParent` or from `rule/operator.go: Match` like I did. Feedback is welcome.

BR,
Thomas


munzzyy:
@b2ag operator.go:340 and operator.go:411 on this branch both handle `OpProcessParentPath`. The second is dead. Master already got that branch in 2509d21e3056, so dropping the block at 340 would keep the diff down to the new operand.

On your ReadCmdline question: at 351 it runs per rule, per connection, walking the whole ancestor chain. If a parent already exited, details.go:327 returns without reading and Args stays empty, so the rule quietly does not match. conman/connection.go:111 and procmon/ebpf/events.go:328 both read the cmdline while the process is alive. Would GetParent() cover your Electron case?

b2ag:
@munzzyy I'm tired of AI slop and the sense of your action really escapes me. You are an AI Slop Energy Vampire and I blocked you.

Closing as it only seems to attract low quality contributions. I use those changes as a patch in a Nix overlay since I posted them here. If it's to niche to go upstream that's fine by me. Bye.

Issue #1318, opened 2025-04-15
[Feature Request] Select connections that don't have destination host name, i.e., that use bare IP address


Andrew15-5:
### Summary:
I want to be able to select requests that use bare IP address instead of host name for destination. I tried `simple` with empty `dest.host` or `regexp` with `^$`, it doesn't work.

Andrew15-5:
After applying the rule

```json
{
  "name": "deny-curl-bare-ip",
  "description": "",
  "action": "deny",
  "duration": "always",
  "operator": {
    "operand": "list",
    "data": "",
    "type": "list",
    "list": [
      {
        "operand": "process.path",
        "data": "/nix/store/k27z9fx6v0msbsl2p54sx4sx0r8247m6-curl-8.12.1-bin/bin/curl",
        "type": "simple",
        "list": null,
        "sensitive": false
      },
      {
        "operand": "dest.host",
        "data": "^$",
        "type": "regexp",
        "list": null,
        "sensitive": false
      }
    ],
    "sensitive": false
  },
  "enabled": true,
  "precedence": false,
  "nolog": false
}
```

I run both `curl https://example.org` and `curl 1.1.1.1`, but they both trigger a new popup. The second should be auto-denied by the rule.

Andrew15-5:
@gustavo-iniguez-goya, I was able to generate a working `flake.nix` to test the thing on NixOS. I tested, and it fails. So I guess this means that it's either a bug, or it's not implemented.

<details><summary>Failing tests</summary>

```sh
go test -C daemon ./rule -test.run='^TestNewOperator.*BareIpNoHostName$'
```

```diff
diff --git a/daemon/rule/operator_test.go b/daemon/rule/operator_test.go
index f7fc27a2..1e39ed60 100644
--- a/daemon/rule/operator_test.go
+++ b/daemon/rule/operator_test.go
@@ -740,3 +740,53 @@ func TestRaceNewOperatorListsDomainsRegexp(t *testing.T) {

 	restoreConnection()
 }
+
+func TestNewOperatorRegexpBareIpNoHostName(t *testing.T) {
+	t.Log("Test NewOperator() regex bare IP (no host name)")
+	var dummyList []Operator
+
+	var sensitive Sensitive = true
+
+	conn.DstHost = ""
+	conn.DstIP = net.ParseIP("1.1.1.1")
+
+	opRE, err := NewOperator(Regexp, sensitive, OpDstHost, "^$", dummyList)
[...]

PR #1398, opened 2025-07-18
Add ability to match bare IP requests (empty host name)

Andrew15-5:
Fixes https://github.com/evilsocket/opensnitch/issues/1318.

The fact that very few relevant things are undocumented makes me do this almost blindly, but from the limited context and knowledge, I do think that empty data field can be present for far more types and operands. But I don't think it makes sense for _all_ types and operands.

### DIFF
--- a/daemon/rule/operator.go
+++ b/daemon/rule/operator.go
@@ -111,8 +111,16 @@ func (o *Operator) Compile() error {
 		return nil
 	}
 
-	// the only operator Type that can have the Data field empty is List.
-	if o.Type != List && o.Operand != OpTrue && o.Data == "" {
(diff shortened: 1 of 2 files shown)


Andrew15-5:
I don't know if [this](https://github.com/evilsocket/opensnitch/actions/runs/16380627268/job/46291169877?pr=1398) is some PR-only workflow, or are tests actually now failing. Because the master branch has the checkmark... Firewall errors sound like it's not the code's fault.

gustavo-iniguez-goya:
don't worry @Andrew15-5 , the tests are failing due to another problem.

thanks!

Issue #1377, opened 2025-06-23
[Bug Report] Fails to pattern-match Appimages with >6 random characters in /tmp/ path


IfGremlinThen:
This is a follow-up to this addition: https://github.com/evilsocket/opensnitch/issues/1066#issuecomment-1826153751

> I've added an option to the pop-ups, to automatically create a regular expression to match the path of AppImages, if the path of the process starts with /tmp/.mount_.

> As far as I can tell, as explained in the commit, the pattern is /tmp/._mount<appId>XXXXXX/.*binary, where only XXXXXX is random. For example IPTVnator -> /tmp/.mount_IPTVnaD8In31/iptvnator, or Siril -> /tmp/.mount_Siril-Cykpw0/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2

As established, most Appimages will randomly generate a /tmp/ path with 6 random characters in the filepath on launch.  OpenSnitch now correctly catches this and provides a prompt with an expression to block outgoing connections with a matching filepath with any 6 random characters:

`^/tmp/\.mount_example\.[0-9A-Za-z]{6}\/.*example$`

This, however, does not stop appimages (or other executables) that create /tmp/ filepaths with more than 6 randomly generated characters such as:

> /tmp/.mount_example.aMxuCxD/usr/bin/example

Because of this OpenSnitch will continually bombard the user with block prompts because the rule it creates does not recognize the filepath.

Editing the expression from a 6 to a 7 (in this example) solves the issue, but it should be expanded to catch any length of randomly generated characters.

`^/tmp/\.mount_example\.[0-9A-Za-z]{7}\/.*example$`

gustavo-iniguez-goya:
Hi @IfGremlinThen ,

The question is, why some appimages are generating paths of 7 random characters, when only 6 is used in the code:
https://github.com/AppImage/AppImageKit/blob/2b8f5b39c41d3895e6b6b88a45b1905f4ffce4ef/src/runtime.c#L523

Can they overwrite what the runtime defines?  hmm, or maybe we're not getting the first part of the path correctly.

Anyway, I'll restrict the regexp to "6 to 7" random characters for now. I'd love to see an example of this issue, maybe you can reproduce it @IfGremlinThen ?

gustavo-iniguez-goya:
Fixed. It will be available on the next release.

IfGremlinThen:
[...]

PR #1399, opened 2025-07-19
Fix: Improve AppImage rule matching by generalizing dynamic mount path

Udityarajsingh:
AppImages often mount to dynamic paths like:

`/tmp/.mount_<random>`

Due to the randomness of these folder names, OpenSnitch was failing to match rules for AppImages unless the mount name remained identical — which it rarely does.

This fix modifies the `_load_rule()` method in the `RulesEditorDialog` to detect AppImage paths starting with `/tmp/.mount_` and replaces them with a generalized regex pattern:

`/tmp/.mount_[A-Za-z0-9]+`

This allows consistent rule matching for AppImages regardless of the specific mount folder name on each launch.

Fixes: #1377

### DIFF
--- a/ui/opensnitch/dialogs/ruleseditor.py
+++ b/ui/opensnitch/dialogs/ruleseditor.py
@@ -470,6 +470,10 @@ class RulesEditorDialog(QtWidgets.QDialog, uic.loadUiType(DIALOG_UI_PATH)[0]):
         self.md5Line.setEnabled(False)
 
     def _load_rule(self, addr=None, rule=None):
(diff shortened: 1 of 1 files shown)


gustavo-iniguez-goya:
Hi @Udityarajsingh !

The issue #1377 is related to the popups dialog. Thanks anyway for looking into it.

Udityarajsingh:
Hi @gustavo-iniguez-goya ,
Thanks for the update!  I'll explore other open issues where I can contribute. Appreciate the clarification!

PR #1401, opened 2025-07-24
getPidFromEbpf new key var typo (nkey)

zel1b08a:
Does this logic-branch make sense by the way?


gustavo-iniguez-goya:
actually no, I think it slipped in during the migration from gobpf to cilium. The fact that the log trace is Error is because I was trying to find some erroneous connections.

Issue #1417, opened 2025-08-25
[Feature Request] allow setting path of audit socket in config


LordGrimmauld:
### Summary:
`audit` will support the `--runstatedir` flag next release, see commit https://github.com/linux-audit/audit-userspace/commit/4ade1466f91dad3c33b2d92596ef8385f3bc5571. In that effort, the default path of the audit socket also changed: Previously, the socket would be placed in `/var/run/audispd_events`, now it will default to `/run/audit/audispd_events`.

Not every system will update `audit` at the same time, and not every system would want to move to the new default path for the audit socket. To accommodate this across all linux distros, i feel it would be appropriate to add a config option to define the path to the socket.

LordGrimmauld:
This is hardcoded in https://github.com/evilsocket/opensnitch/blob/1e3893997c56a9c1b6ca240f4c91e0549472b5f7/daemon/procmon/audit/client.go#L98, and would need a config value similar to how we already have `ModulesPath` for ebpf.

LordGrimmauld:
@gustavo-iniguez-goya audit 4.1.2 with the new socket path has now released: https://github.com/linux-audit/audit-userspace/releases/tag/v4.1.2

I don't know enough Go to finalize my PR

gustavo-iniguez-goya:
no problem @LordGrimmauld , I'll take a look at it.

PR #1418, opened 2025-08-26
daemon/audit: allow configuring audispd_events socket path

LordGrimmauld:
Closes https://github.com/evilsocket/opensnitch/issues/1417

Draft until i did my testing.


LordGrimmauld:
Hmm, i am a little confused about why the test CI is failing, i don't see how validate-invalid-config breaks. I tried to replicate the pattern of how ebpf does config parsing, but apparently i am missing something?

gustavo-iniguez-goya:
The tests were failing because there's a comma missing in the file `daemon/ui/testdata/config-invalid-procmon.json`.

On the other hand, it seems that `auditctl -A ...` fails if there're no rules already added. I don't remember having this problem in the past, but probably `auditctl` has changed how it works.

Anyway, thanks @LordGrimmauld !

LordGrimmauld:
oh oops i am dumb, indeed that missing comma is problematic... But thank you for looking into this, this config will be helpful :)

Issue #1431, opened 2025-09-15
[Bug Report] Forced Full-Screen


cs50ad:
### Describe the bug:
The recent update (1.7.2) has caused OpenSnitch to go full-screen - over my desktop panel, and without a menu/titlebar. I have tried the usual shortcuts to exit (Esc, Alt + F11, etc.). I've also tried to move the window using a keybinding, but it seems to be an immovable window. The only way to exit is if I maximize a different app, which brings up my desktop panel, and then exiting out of OpenSnitch's GUI through its panel icon.

Important information:
 - OpenSnitch version: 1.7.2
 - OS: Linux Mint 22.2 Cinnamon 6.4.8
 - Kernel version: 6.14.0-29-generic

### To Reproduce:
1. Open OpenSnitch's GUI
2. The GUI will open as full-screen

### Expected behavior:
The GUI should behave as it has before this update.

### Additional context:
I installed the app through the provided .deb files. I am also using X11, not Wayland.

Is there a command I can enter or something in the config file that I can add to fix this issue?
Thank you very much for this wonderful app, and thank you for any help that may be provided.

x80286:
having this same issue..

wb4r:
same ^

Copyilot:
Same

soda-pop-ice-cream:
Having same issue, also Cinnamon, X11, and Linux Mint.

I just Alt+F4 this window, so opensnitch-ui running in system tray anyway

ryan-steed-usa:
A demonstrative screenshot might help diagnosis. Have you consulted the [wiki](https://github.com/evilsocket/opensnitch/wiki/GUI-known-problems#gui-size-problems-on-4k-monitors)? Could it be a scaling problem? There should be a GUI config if you're on a recent version:

<img width="652" height="353" alt="Image" src="https://github.com/user-attachments/assets/48336827-f838-4689-8ca0-6762870922e7" />

I couldn't reproduce this behavior using a _Linux Mint 22.2 Cinnamon 6.4.8_ virtual machine. If scaling isn't the problem, you could try forcing the `xcb` _Qt platform plugin_ in the GUI options. Otherwise, you might try using a terminal to execute `opensnitch-ui` manually to reveal its standard output which may contain a descriptive error.

Environment variables as described in the wiki can be set to temporarially control these options. Or the config can be manually edited, the relevant values might something like this (under `[global]`):
```
[global]
qt_platform_plugin=xcb
[...]

Issue #1374, opened 2025-06-16
[Bug Report] GUI main window won't remember window diminsions in xcb mode


user10072023github:
Hi Gustavo, love the proggy so much. Minor issue.

### Describe the bug:
The main GUI window won't remember dimensions when closed and reopened.

 - OpenSnitch version: 1.7.0.0 latest
 - OS: Fedora
 - OS version: 42 Workstation
 - Window Manager: GNOME Wayland
 - Kernel version: 6.14.9-300

### To Reproduce:
Install Fedora 42 Workstation w/ default settings
Install OpenSnitch daemon and GUI via RPM packages
Install additional themes via "pip3 install qt-material
Reboot
Open the main GUI window and resize to desired dimensions, e.g. fullscreen
Close the main GUI window and reopen

### Expected behavior (optional):
The main GUI window would hopefully recall what dimensions the window had at prior exit and whether it was fullscreen or not.

### Workaround:
Select the "Maximize" button on the window when reopening.

gustavo-iniguez-goya:
As far as I can tell, this issue only occurs when the window is maximized/fullscreen and it's closed. I'll take a look at it.

gustavo-iniguez-goya:
Fixed. I've had some issues setting window's states on Wayland, so I've decided to resize the window to the screen geometry.
Only tested on Cinnamon and Gnome for now.

user10072023github:
So appreciated!

PR #1458, opened 2025-11-17
fix: restore maximized window state correctly when loading dialog geometry

ryan-steed-usa:
This change fixes an issue where the statistics dialog would not properly restore its maximized state when reopening. The previous implementation attempted to manually set the window geometry to simulate a maximized state, but this approach was unreliable and didn't properly handle screen boundaries or multi-monitor setups.

The fix replaces the manual geometry setting with a proper call to showMaximized() when the dialog is visible, ensuring that Qt's window management correctly handles the maximized state restoration. This provides consistent behavior across different platforms and display configurations.

Changes made:
- Removed the manual geometry calculation that attempted to simulate maximized state
- Added proper isVisible() check before calling showMaximized() to ensure the window is properly initialized
- Simplified the maximized state restoration logic to use Qt's built-in functionality

A potential caveat is that [showMaximized()](https://doc.qt.io/qt-6/qwidget.html#showMaximized) may [encounter problems with X11](https://doc.qt.io/qt-6/application-windows.html#x11-peculiarities) although the benefit may outweigh the risk to enhance multi-monitor capability and better compensate for desktop resolution changes. More testing is warranted. Patched 1.7.2 seems to work as expected with **KDE Plasma 6.5.2 with Wayland**, **Linux Mint 22.2 Cinnamon 6.4.8**, **Ubuntu 24.04 with Gnome 46** and **Ubuntu 25.04 with Gnome 48**.

_X11 provides no standard or easy way to get the frame geometry once the window is decorated. Qt solves this problem with nifty heuristics and clever code that works on a wide range of window managers that exist today. Don't be surprised if you find one where [QWidget::frameGeometry](https://doc.qt.io/qt-6/qwidget.html#frameGeometry-prop)() returns wrong results though._

_Nor does X11 provide a way to maximize a window. [QWidget::showMaximized](https://doc.qt.io/qt-6/qwidget.html#showMaximized)() has to emulate the feature. Its result depends on the result of [QWidget::frameGeometry](https://doc.qt.io/qt-6/qwidget.html#frameGeometry-prop)() and the capability of the window manager to do proper window placement, neither of which can be guaranteed._

### DIFF
--- a/ui/opensnitch/dialogs/stats.py
+++ b/ui/opensnitch/dialogs/stats.py
@@ -849,8 +849,8 @@ class StatsDialog(QtWidgets.QDialog, uic.loadUiType(DIALOG_UI_PATH)[0]):
                     # docs: https://doc.qt.io/qt-5/qwidget.html#setWindowState
                     self.setWindowState(self.windowState() & ~QtCore.Qt.WindowState.WindowMinimized | QtCore.Qt.WindowState.WindowActive)
 
-    def showEvent(self, event):
-        super(StatsDialog, self).showEvent(event)
+    def show(self):
+        super(StatsDialog, self).show()
         self._shown_trigger.emit()
         window_title = QC.translate("stats", "OpenSnitch Network Statistics {0}").format(version)
         if self._address is not None:


ryan-steed-usa:
Closes #1431

gustavo-iniguez-goya:
thank you @ryan-steed-usa !

I didn't use showMaximized() because it didn't work well on Xwayland #1374, I don't remember the problem but I think it didn't restore the state correctly on Wayland.
Anyway, I've tested the PR on Xorg and Xwayland. On Xorg seems to work fine, but on Wayland (Arch, updated) sometimes when clicking on the systray icon to restore the GUI to the maximized state it fails with:

```
dialogs/stats.py, line 867, eventFilter
RecursionError: maximum recursion depth exceeded
```

The problem is that showMaximized() fires the event showEvent(), in showEvent() we load again the settings, showMaximized() is called again, and we end up in a loop.

One option could be to not override showEvent(), and replace it with show() (I've only tested it on Arch/Wayland and Mint22/Xorg):

def showEvent() -> def show()

```
def show(self):
  ...
  self.load_settings()
  ...
  super(StatsDialog, self).show()
```

See please if you can reproduce this bug.

ryan-steed-usa:
@gustavo-iniguez-goya

> See please if you can reproduce this bug.

Forgive me. I'd only tested with manually patched **1.7.2** versions, where I cannot reproduce the bug. But I have successfully reproduced the recursion error with master branch builds (**1.7.3** candidate) while testing with Archlinux and Ubuntu. I'm guessing that the PyQT6 migration, or Qt6 itself, has manifest the bug?

Per your suggestion, I've added a new commit that does resolve the error everywhere I've tested. Let me know if it's acceptable or not. Thanks!

gustavo-iniguez-goya:
thank you @ryan-steed-usa ! just in time for the next release.

PR #1480, opened 2025-12-20
tests: port ui tests to qt6

themighty1:
I noticed that after the port to Qt6 the tests were not ported, so I ported them. What i did:

- port UI tests to Qt6
- add new tests where it makes sense
- refactored to use direct button clicks instead of QTimer.singleShot() + exec(). This is the recommended pattern in the pytest-qt community for testing dialogs - it avoids fragile timing dependencies and reduces test runtime from ~15s to ~2.4s.

Tests can be run from ui/ with:
> QT_QPA_PLATFORM=offscreen pytest -v tests/

Also while testing 2 bugs were discovered, I'll inline them here, feel free to fix them 

<details>
  <summary>Bug 1</summary>

The Bug:

  In opensnitch/dialogs/ruleseditor/utils.py:

  - Line 143 calls win._is_valid_regex(sp_regex) - treating it as a method on the dialog object
  - Line 181 defines is_valid_regex(win, regex) - as a module-level function, not a method

  So when you enter comma-separated ports like "80,443,8080", the code in comma_to_regexp() tries to validate the generated regex by calling win._is_valid_regex(), but that method doesn't exist on RulesEditorDialog. The correct call should be:

  Line 143 currently says:
  if not win._is_valid_regex(sp_regex):

  Should be:
  if not is_valid_regex(win, sp_regex):

  Impact: Any rule using comma-separated values (ports, IPs, hosts, protocols, interfaces) will fail to save because the validation crashes before the rule can be created.

  The fix is a one-line change - call the module function instead of a non-existent method. Want me to fix it?

  </details>

  <details>
  <summary>Maybe bug 2</summary>

in ui/opensnitch/nodes.py:delete_all(): Sets self._nodes = None instead of self._nodes = {}. This causes AttributeError when other methods (e.g., get(), count(), get_nodes()) access _nodes.keys() or iterate over it.

  Fix:
  def delete_all(self):
      self.stop_notifications()  # Send exit signal to all nodes
      self._nodes = {}
      self.nodesUpdated.emit(self.count())

  </details>

### DIFF
--- a/ui/tests/conftest.py
+++ b/ui/tests/conftest.py
@@ -1,0 +1,77 @@
+# conftest.py - pytest configuration for opensnitch UI tests
+#
+# This file sets up Qt and database before tests run.
+
+import pytest
+from PyQt6 import QtWidgets
+from unittest.mock import patch
+from queue import Queue
+
+# Global flag to track initialization
+_initialized = False
+
+def init_test_environment():
+    """Initialize database and config after QApplication exists."""
+    global _initialized
+    if _initialized:
+        return
+
+    from opensnitch.database import Database
+    from opensnitch.config import Config
+    from opensnitch.nodes import Nodes
+
+    db = Database.instance()
+    db.initialize()
+    Config.init()
+
+    # Setup mock node with full structure
+    from tests.dialogs import ClientConfig
+    nodes = Nodes.instance()
+    nodes._nodes["unix:/tmp/osui.sock"] = {
+        'data': ClientConfig,
+        'notifications': Queue(),
+        'online': True
+    }
+
+    _initialized = True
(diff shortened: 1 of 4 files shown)


gustavo-iniguez-goya:
thank youu @themighty1 ! :raised_hands: 

> The fix is a one-line change - call the module function instead of a non-existent method. Want me to fix it?

Ouch, what a mistake. fixed.

Regarding the other issue, it seems that I fixed it while doing other changes:
https://github.com/evilsocket/opensnitch/blob/02aab314c0c9912d7eb4f1bbec8b7191c46198a6/ui/opensnitch/nodes.py#L145-L148

PR #1491, opened 2026-01-01
ui,nodes: fixed delete_all() sending invalid notification

themighty1:
This PR fixes a bug (called Bug 2 in https://github.com/evilsocket/opensnitch/pull/1480) 
I think there was some miscommunication and the bug was left unfixed.

Also removes the test workaround needed due to the bug.


gustavo-iniguez-goya:
ops, you're right. thanks !

PR #1493, opened 2026-01-02
tests: added netfilter tests

themighty1:
This PR adds integration tests for the netfilter functionality.

  - Added netfilter integration tests
  - Refactored iptables rule building into a dedicated function to eliminate code duplication between production and test code
  - Extracted network test harness into a shared module used by both eBPF and netfilter tests

  The production code was modified to extract rule-building logic into a testable function. This approach maintains a single source of truth, but I'm open to splitting the production changes into a separate PR if preferred.

### DIFF
--- a/daemon/firewall/iptables/rules.go
+++ b/daemon/firewall/iptables/rules.go
@@ -16,6 +16,37 @@ func (ipt *Iptables) getBypassQueue() string {
 	return "--queue-bypass"
 }
 
+// BuildQueueDNSRule returns the iptables rule arguments for queueing DNS responses.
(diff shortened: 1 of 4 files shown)


gustavo-iniguez-goya:
perfect :ok_hand: !

By the way, when we added OpenSnitch to Debian I disabled some tests that require kernel level privileges, because they cannot run on their tests environment, which are chroots or LXC containers. 
I added PRIVILEGED_TESTS= to skip these tests, disabled by default. We'll run into the same issues with these tests that require access to the kernel if there's no way to disable them.

themighty1:
That's a good observation, thanks. 
I will keep that in mind and also will check Debian testing procedures.

PR #1495, opened 2026-01-04
fix(ui): prevent deadlock when approving NFS connections (#1334)

### LINKED ISSUE #1334: [Bug Report] Deadlock when asked to approve Kernel Connection for NFS with home directory
When using eBPF based OpenSnitch 1.6.8 (package from Debian), on a workstation with user home on a NFS server, the popup to approve the kernel NFS server connection seem to deadlock when approving the connection.  I suspect the client try to write to the NFS volume before the kernel has been told to approve the connection.  The popup stay up after I press 'Accept', and the only way to get further is to do a simple `pkill -15 opensnitch-ui` and restart the GUI using the desktop menu entry.

I could find nothing sensible in the logs. :(

po-nuvai:
## Summary

Fixes #1334 - Deadlock when asked to approve Kernel Connection for NFS with home directory

When the user's home directory is on NFS and a kernel NFS connection needs approval, a synchronous disk write to save dialog geometry would cause a deadlock:

1. Kernel NFS connection arrives and blocks waiting for approval
2. User clicks "Allow" in the prompt dialog
3. `_send_rule()` calls `setSettings("promptDialog/geometry", ...)` which performs a synchronous disk write to `~/.config/opensnitch/`
4. The disk write blocks because it needs NFS (which is waiting for approval)
5. **Deadlock**: The connection can't get approval until the rule is returned, but the UI is blocked on NFS

### Changes

- Added `_save_geometry()` helper method that wraps the settings write
- Removed synchronous geometry save from the beginning of `_send_rule()`
- Added deferred geometry save using `QTimer.singleShot(0, ...)` in the `finally` block
- The deferred call runs after `self._done.set()`, ensuring the rule is returned to the daemon first

This breaks the circular dependency by ensuring the NFS connection approval is sent before any disk writes that depend on NFS.

## Test Plan

- [ ] Test on a system with home directory mounted via NFS
- [ ] Approve a kernel NFS connection - should not deadlock
- [ ] Verify dialog geometry is still saved correctly after approval
- [ ] Test normal (non-NFS) connections still work as expected

## Reviewers

@petterreinholdtsen - Original reporter who experienced the issue

### DIFF
--- a/ui/opensnitch/dialogs/prompt/dialog.py
+++ b/ui/opensnitch/dialogs/prompt/dialog.py
@@ -579,9 +579,21 @@ class PromptDialog(QtWidgets.QDialog, uic.loadUiType(DIALOG_UI_PATH)[0]):
             self.checkDstIP.isChecked() or \
             self.checkSum.isChecked()
 
-    def _send_rule(self):
+    def _save_geometry(self):
+        """Save dialog geometry to settings.
+
+        This is called via QTimer.singleShot to avoid blocking the rule
+        approval flow. When the home directory is on NFS and the NFS
+        connection itself needs approval, a synchronous disk write would
+        cause a deadlock (issue #1334).
+        """
         try:
             self._cfg.setSettings("promptDialog/geometry", self.saveGeometry())
+        except Exception as e:
+            self.logger.debug("Failed to save prompt dialog geometry: %s", repr(e))
+
+    def _send_rule(self):
+        try:
             self._rule = ui_pb2.Rule(name="user.choice")
             self._rule.created = int(datetime.now().timestamp())
             self._rule.enabled = True
@@ -690,3 +702,6 @@ class PromptDialog(QtWidgets.QDialog, uic.loadUiType(DIALOG_UI_PATH)[0]):
             # a new rule is available
             self._done.set()
             self.hide()
+            # Defer geometry save to avoid blocking the rule approval flow.
+            # This prevents deadlock when home directory is on NFS (issue #1334).
+            QtCore.QTimer.singleShot(0, self._save_geometry)


petterreinholdtsen:
[Pothihai Selvan]
> @petterreinholdtsen - Original reporter who experienced the issue

I would love to test this, but lack the needed Go compiler knowledge to
get newer versions to build.  I've tried a few times to update the
Debian packages the last few weeks, without success.

-- 
Happy hacking
Petter Reinholdtsen

po-nuvai:
@petterreinholdtsen Thanks for your willingness to test! 

Good news - this particular fix is actually in the **Python/Qt UI code** (`ui/opensnitch/dialogs/prompt/dialog.py`), not the Go daemon. So you might be able to test it without needing to compile anything!

**Quick test option:**

1. Back up your current file:
   ```bash
   sudo cp /usr/lib/python3/dist-packages/opensnitch/dialogs/prompt/dialog.py \
           /usr/lib/python3/dist-packages/opensnitch/dialogs/prompt/dialog.py.bak
   ```

2. Download just the patched file from this PR:
   ```bash
   sudo curl -o /usr/lib/python3/dist-packages/opensnitch/dialogs/prompt/dialog.py \
        https://raw.githubusercontent.com/po-nuvai/opensnitch/fix/nfs-deadlock-1334/ui/opensnitch/dialogs/prompt/dialog.py
   ```

3. Restart the UI:
   ```bash
   opensnitch-ui &
   ```

4. Trigger an NFS-related connection prompt and see if it still deadlocks.

To revert: just copy back the `.bak` file and restart the UI.

Let me know if this works for your setup or if you need any help! 🙂

PR #1498, opened 2026-01-04
fix(ui): make table cells read-only to prevent accidental editing (#1285)

### LINKED ISSUE #1285: [Bug Report] Double-clicking table fields makes them editable, which doesn't make sense
### Describe the bug:
In most tabs, double-clicking table fields makes them editable. That's useful to copy text, but would it be possible to make the cells read-only?

Include the following information:
 - OpenSnitch version: 1.6.8
 - Window Manager: KDE

### To Reproduce:

1. Go to 'Applications'
2. Double click on a cell in the 'Time' column

Image

po-nuvai:
## Summary

Fixes #1285 - Double-clicking table fields makes them editable, which doesn't make sense

The issue reported that double-clicking on cells in tables (Events, Hosts, Applications, Users, Rules tabs) would make them editable, entering an edit mode even though changes are not saved. This created a confusing user experience.

### Changes

- Added `setEditTriggers(QTableView.EditTrigger.NoEditTriggers)` in `GenericTableView.__init__()` to disable all edit triggers
- This affects all tables that inherit from GenericTableView (Events, Hosts, Applications, Users, Rules, Addresses, Ports, and Procs tabs)
- Text selection for copying is still possible via row selection

### Technical Details

`GenericTableView` is the base class for all the main data tables in the OpenSnitch UI. By setting `NoEditTriggers` at the base class level, all derived tables automatically become read-only without needing individual modifications.

## Test Plan

- [ ] Double-click on cells in Events tab - should not enter edit mode
- [ ] Double-click on cells in Hosts tab - should not enter edit mode  
- [ ] Double-click on cells in Applications tab - should not enter edit mode
- [ ] Double-click on cells in Users tab - should not enter edit mode
- [ ] Double-click on cells in Rules tab - should not enter edit mode
- [ ] Verify row selection still works for copying data

### DIFF
--- a/ui/opensnitch/customwidgets/generictableview.py
+++ b/ui/opensnitch/customwidgets/generictableview.py
@@ -228,6 +228,9 @@ class GenericTableView(QTableView):
         self.horizontalHeader().setStretchLastSection(True)
         #the built-in vertical scrollBar of this view is always off
         self.setVerticalScrollBarPolicy(Qt.ScrollBarPolicy.ScrollBarAlwaysOff)
+        # Make cells read-only to prevent accidental editing on double-click (#1285).
+        # Text selection for copying is still possible via row selection.
+        self.setEditTriggers(QTableView.EditTrigger.NoEditTriggers)
         self.installEventFilter(self)
 
     def setVerticalScrollBar(self, vScrollBar):


gustavo-iniguez-goya:
This is the actual expected behaviour.

PR #1500, opened 2026-01-04
fix(ui): use temporary duration when popup times out (#1385)

### LINKED ISSUE #1385: [Bug Report] Expiring advanced view pop-up will create blanket deny rule for application of "Destianion IP" is not checked
### Describe the bug:

This is a follow-up to #1368.

 - OpenSnitch version: 1.7.0
 - OS: Ubuntu 25.04
 - Window Manager: KDE 6
 - Kernel version: `6.14.0-22-generic #22-Ubuntu SMP PREEMPT_DYNAMIC Wed May 21 15:01:51 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux`

### To Reproduce:

1. Configure the Pop-up defaults to deny, forever, by executable, and in `More`: `Show advanced view by default`, and leave `Destination IP` unchecked (default)
2. Have an application connect to a new domain (no rule matches it)
3. Let the pop-up timer expire. The pop-up will close.
4. Under Rules -> Application rules -> Permanent, notice the new `deny` rule for the application, regardless of IP.

### Expected behavior (optional):

The rule should be temporary, "once", not permanent.

OpenSnitch creating a permanent `deny` rule for the executable regardless of the IP is a major problem if the pop-up is displayed when the machine is unattended. As a workaround, `Destination IP` can be checked, but this setting isn't easily discoverable.

### Screencast

https://github.com/user-attachments/assets/34b15590-5198-4a69-9efc-5285154696a4

po-nuvai:
## Summary

Fixes #1385 - Expiring advanced view pop-up will create blanket deny rule for application if "Destination IP" is not checked

### The Problem

When a popup expires (times out) without user interaction, it currently applies all the configured default settings, including the duration. If the user has "forever" as the default duration, this creates **permanent blanket deny rules** when the user is away from the machine.

This is dangerous behavior for a security application - a user who steps away shouldn't return to find their applications permanently blocked.

### The Fix

When `_timeout_triggered` is True, the duration is now forced to `"once"` instead of using the configured default:

```python
if self._timeout_triggered:
    self._rule.duration = Config.DURATION_ONCE
else:
    self._rule.duration = utils.get_duration(self.durationCombo.currentIndex())
```

### Behavior Change

| Scenario | Before | After |
|----------|--------|-------|
| User clicks Allow/Deny | Uses selected duration | Uses selected duration (unchanged) |
| Popup times out | Uses default duration (could be "forever") | Forces "once" duration |

### Rationale

- **Explicit confirmation = full control**: When the user actively clicks, they're confirming their choice including the duration
- **Timeout = conservative default**: When the popup expires unattended, the system shouldn't make permanent decisions
- **Desktop notifications still inform**: The user will see a notification about the temporary rule and can review/make permanent in the Rules tab

## Test Plan

- [ ] Set default duration to "forever" and default action to "deny"
- [ ] Trigger a popup for a new connection
- [ ] Let the popup expire without clicking
- [ ] Verify the created rule has duration "once" (not "forever")
- [ ] Verify actively clicking Deny creates a "forever" rule
- [ ] Verify notification is still shown for the timeout rule

### DIFF
--- a/ui/opensnitch/dialogs/prompt/dialog.py
+++ b/ui/opensnitch/dialogs/prompt/dialog.py
@@ -585,7 +585,13 @@ class PromptDialog(QtWidgets.QDialog, uic.loadUiType(DIALOG_UI_PATH)[0]):
             self._rule = ui_pb2.Rule(name="user.choice")
             self._rule.created = int(datetime.now().timestamp())
             self._rule.enabled = True
-            self._rule.duration = utils.get_duration(self.durationCombo.currentIndex())
+            # If the popup timed out without user interaction, always use "once" duration
+            # to avoid creating permanent rules when the user is away (#1385).
+            # Only explicit user confirmation should create permanent rules.
+            if self._timeout_triggered:
+                self._rule.duration = Config.DURATION_ONCE
+            else:
+                self._rule.duration = utils.get_duration(self.durationCombo.currentIndex())
 
             self._rule.action = Config.ACTION_ALLOW
             if self._default_action == Config.ACTION_DENY_IDX:


gustavo-iniguez-goya:
There's a discussion open, and this is not a bug, it's the intended behaviour.
On the other hand the behaviour is configurable by the user, and since some versions ago the default duration is 12h.

Issue #1341, opened 2025-05-06
[Bug Report] Only one user can use OpenSnitch GUI at the same time


Damglador:
### Describe the bug:
For example, I have user 1 on tty1 and user 2 on tty2, if OpenSnitch GUI is running for user 1, user 2 can't have it, it simply doesn't launch.

This will be an issue if a user logs, gets OpenSnitch GUI and then just locks the session without quitting. It also might be an issue with multiset setups, but that's just a guess.

Include the following information:
 - OpenSnitch version: 1.6.8
 - OS: ArchLinux
 - Window Manager: Plasma
 - Kernel version: Linux Parasite 6.14.5-2-cachyos

### To Reproduce:
1. Log as user 1 on tty1
2. Start a graphical session
3. Start OpenSnitch GUI
4. Switch to tty2
5. Log as user2
6. Start a graphical session
7. Start OpenSnitch GUI
8. It doens't start

### Post error logs:
```
WARNING:root:qt_material must be imported after PySide or PyQt!
         ~ OpenSnitch GUI - 1.6.8 ~
        protobuf: 6.30.2 - grpc: 1.72.0
-------------------------------------------------- 

QT_AUTO_SCREEN_SCALE_FACTOR: True
gRPC Max Message Length: None
                  Bytes: 4194304
is new file, or IN MEMORY, setting initial schema version
setting schema version to: 3
setting schema version to: 3
Setting journal_mode:  OFF
Setting DB memory optimizations
schema version: 3
db schema is up to date
Loading translations: /usr/lib/python3.13/site-packages/opensnitch/utils/../i18n locale: uk_UA
using IPASN DB: /usr/lib/python3/dist-packages/data/ipasn_20140513_v12.dat.gz
exception loading ipasn db: [Errno 2] Немає такого файла або каталогу: '/usr/lib/python3/dist-packages/data/ipasn_20140513_v12.dat.gz'
Install python3-pyasn to display IP's network name.
Using server address: unix:///tmp/osui.sock auth type: simple
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
E0000 00:00:1746557172.825165  288475 chttp2_server.cc:1685] UNKNOWN:No address added out of total 1 resolved for 'unix:///tmp/osui.sock' {children:[UNKNOWN:Unable to configure socket {children:[UNKNOWN:bind: Адресу вже використано (98)]}]} (translation:  adress is already used)
[...]

PR #1501, opened 2026-01-04
ui: inform about failure to connect to socket

Damglador:
Not quite a fix for https://github.com/evilsocket/opensnitch/issues/1341, but at least it makes it clear for the user that the issue is in another opensnitch-ui instance and is not some random bug.

Instead of silently crashing, opensnitch-ui will pup-up the dialog shown on the screenshot and only then exit:
Знімок_20260104_170731


gustavo-iniguez-goya:
thanks @Damglador !

PR #1513, opened 2026-01-24
Match by full command by default for high-risk executables

ngvuhuy:
The default behavior when creating a rule using UI popup is matching by executables. This, however, is too broad for apps and scripts built with Python and node, or programs like `curl` and `wget`, because allowing `python3` to access a website means every scripts ran using `python3 script.py` will have access to that website.

This problem is [discussed previously](https://github.com/evilsocket/opensnitch/discussions/612#discussioncomment-2116878) and [included in Wiki](https://github.com/evilsocket/opensnitch/wiki/Rules-examples#filtering-python-scripts-applicable-to-java-and-others-interpreters), but is not widely known. This PR fixes that by setting the rule to match the entire command line by default, including all arguments, for such high-risk executable.

Screenshot_20260124_235622

Currently, the matched executables are `curl`, `wget`, `python`, `node`, `java`, `ssh`

Programs that fall into this bucket are much more common than people actually expect, even if they do not frequently program in those languages and/or use the command line. For example, this includes Lutris, Bottles, Anki, Claude Code, any `curl | bash` installation process, and Git over SSH.

### DIFF
--- a/ui/opensnitch/config.py
+++ b/ui/opensnitch/config.py
@@ -111,6 +111,7 @@ class Config:
     DEFAULT_POPUP_ADVANCED_CHECKSUM = "global/default_popup_advanced_checksum"
     DEFAULT_FW_INTERCEPTION_ENABLED = "global/interception_enabled"
     DEFAULT_PERSIST_INTERCEPTION_STATE = "global/persist_interception_state"
+    DEFAULT_POPUP_ADVANCED_FULL_COMMAND = "global/default_popup_advanced_full_command"
     DEFAULT_SERVER_ADDR  = "global/server_address"
     DEFAULT_SERVER_MAX_MESSAGE_LENGTH  = "global/server_max_message_length"
     DEFAULT_SERVER_MAX_WORKERS = "global/max_workers"
--- a/ui/opensnitch/dialogs/preferences/settings.py
+++ b/ui/opensnitch/dialogs/preferences/settings.py
@@ -9,6 +9,7 @@ ui_pb2, ui_pb2_grpc = proto.import_()
 from opensnitch.config import Config
 from opensnitch.utils import languages
 from opensnitch.database import Database
+from opensnitch.dialogs.prompt import constants as prompt_constants
 from opensnitch.dialogs.preferences import (
     utils,
 )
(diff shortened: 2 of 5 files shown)


gustavo-iniguez-goya:
hi @blaccod !

I agree that allowing interpreters can be dangerous. But we cannot hard-code the list of commands without giving the user the ability to configure it.

In any case, I'd not include in that list curl, wget or ssh. If we include those binaries, we should also include bash, awk, rsync, smbclient, ftp, etc, etc. (or the whole gtfobin list).

What if we add this PR, and allow users to configure the list of commands that will preselect path + cmdline?

ngvuhuy:
> What if we add this PR, and allow users to configure the list of commands that will preselect path + cmdline?

That sounds good, I can add it under Preferences > Rules (or Prefereces > Popup > More) and note that it is an advanced settings, and have a small hover tooltip to explain what that is.
I'm not entirely sure how the input format should be:
- A comma-separated list of executable name, similar to what's in the code of this PR.
- A regex matching the same thing (e.g. `python.*|node.*`)
- A path to text file containing full paths of all high-risk executables

The reason this PR has to do the `pathlib.Path` thing instead of a simple `.endswith()` check is because the distros and Flatpak usually ship different versions of the same interpreters under different paths (e.g. `python` also has `python3.11`/`python3.12` or `node`/`node24`), and expecting us/the users to update this list every new version update is a bit much. Therefore I'm leaning more towards the regex solution (and also have a good default regex included)

> In any case, I'd not include in that list curl, wget or ssh.

I can't think of any time when allowing curl and wget to indiscriminately connect to the Internet is a good thing, that's why they're there. ssh is there because `git clone` uses it under the hood, but I agree that is not as high-risk to be included here, and we should let the user set that themselves instead.

gustavo-iniguez-goya:
> That sounds good, I can add it under Preferences > Rules (or Prefereces > Popup > More) and note that it is an advanced settings, and have a small hover tooltip to explain what that is.

perfect :+1:  All the pop-ups related configuration is under the Pop-up tab. Maybe it'd be better to add it there, for consistency.

> A comma-separated list of executable name, similar to what's in the code of this PR.

I'd go for this. It's what you already did, and it's configurable. If someone has other necessity they can open a feature request explaining their need.

The list should be saved/loaded using the load_ui_settings() and save_ui_settings().

> I can't think of any time when allowing curl and wget to indiscriminately connect to the Internet is a good thing, that's why they're there. ssh is there because git clone uses it under the hood, but I agree that is not as high-risk to be included here, and we should let the user set that themselves instead.

Bear in mind that users can create rules to filter by multiple fields, so for example you can create a rule to allow `/usr/bin/curl` to connect to port 443 and IP `192.168.1.*`, regardless of the cmdline. Or a more common scenario: allow `/usr/bin/ssh` port 22, dst: LAN.

Filtering by default by cmdline can cause "pop-ups fatigue", as someone has already defined it.

If it's a server, honestly I'd not even install wget or curl if I don't explicitly need it, given how commonly they're used in attacks.

ngvuhuy:
Finally have some time to sit down and finish the UI config
image
Now we can pass in a list of comma-separated command names and any command that matches the start of an entry on that list would default to filter by command line in pop-ups

PR #1531, opened 2026-02-13
fix(ui): null-check screen in PromptDialog

staticssleever668:
Fixes crash when pressing 'accept' with keyboard when mouse cursor is not on top of the popup.

> Traceback (most recent call last):
>   File "/path/to/opensnitch/dialogs/prompt/dialog.py", line 288, in _check_advanced_toggled
>     self.move_popup()
>     ~~~~~~~~~~~~~~~^^
>   File "/path/to/opensnitch/dialogs/prompt/dialog.py", line 254, in move_popup
>     point = self.screen().virtualSiblingAt(QtGui.QCursor.pos()).availableGeometry()
>             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'NoneType' object has no attribute 'availableGeometry'
> Fatal Python error: Aborted


gustavo-iniguez-goya:
thanks!

Issue #1537, opened 2026-02-15
[Bug Report] eBPF Module Fails To Build Against CachyOS's 6.19 Kernel


Joomzie:
### Describe the bug:  

This may only affect CachyOS, as Arch's owm `linux` package hasn't been updated to 6.19 yet, but the eBPF module fails to build when using the `opensnitch-ebpf-module-git` AUR package. I reported this problem there as well just in case this isn't an upstream issue.

 - OpenSnitch version: 1.8.0.r125.8bc2698d-1
 - OS: CachyOS
 - OS version: Rolling
 - Window Manager: COSMIC
 - Kernel version: 6.19.0-2-cachyos

### To Reproduce:  

Attempt to build the eBPF module using the PKGBUILD from the AUR.

### Post error logs:  

```
clang -I. -I/usr/lib/modules/6.19.0-2-cachyos/build/arch/x86/include/ -I/usr/lib/modules/6.19.0-2-cachyos/build/arch/x86/include/generated/ -I/usr/lib/modules/6.19.0-2-cachyos/build/include -include /usr/lib/modules/6.19.0-2-cachyos/build/include/linux/kconfig.h -I/usr/lib/modules/6.19.0-2-cachyos/build/include -I/usr/lib/modules/6.19.0-2-cachyos/build/include/uapi -I/usr/lib/modules/6.19.0-2-cachyos/build/include/generated/uapi -I/usr/lib/modules/6.19.0-2-cachyos/build/arch/x86/include -I/usr/lib/modules/6.19.0-2-cachyos/build/arch/x86/include/generated -I/usr/lib/modules/6.19.0-2-cachyos/build/arch/x86/include/uapi -I/usr/lib/modules/6.19.0-2-cachyos/build/arch/x86/include/generated/uapi -I/usr/lib/modules/6.19.0-2-cachyos/build/tools/testing/selftests/bpf/ -D__KERNEL__ -D__BPF_TRACING__ -Wno-unused-value -Wno-pointer-sign -D__TARGET_ARCH_x86 -Wno-compare-distinct-pointer-types  -Wunused -Wno-unused-value -Wno-gnu-variable-sized-type-not-at-end -Wno-address-of-packed-member -Wno-tautological-compare -Wno-unknown-warning-option -fcf-protection -fno-stack-protector -g -O2 -emit-llvm -c opensnitch.c
[...]

PR #1554, opened 2026-02-27
fix kernel 6.19 build, add some hardening options

munix9:
solves #1537


gustavo-iniguez-goya:
thanks @munix9 ! 

I doubt that we need native compiler hardening features. Those validations are for machine code, not eBPF bytecode.
The eBPF VM is who checks in this case boundaries, branches and other validations.

PR #1560, opened 2026-03-07
persist firewall interception state across restarts

zackslash:
Hi,

When the OpenSnitch GUI is restarted, the pause state is always lost. The firewall starts as "running" regardless of whether the user had it paused before closing. Users who routinely pause the firewall on startup must manually re-pause every time.

This MR adds an opt-in preference to persist the firewall interception (pause) state across GUI restarts.

When enabled, the GUI remembers if the user had paused the firewall and automatically restores that state on next startup.

The new option "remember pause state across restarts"  was added under 'general' preferences and defaults to off to preserve the original behavior.

Screenshot_2026-03-07_14-42-53


gustavo-iniguez-goya:
thank you @zackslash !

PR #1565, opened 2026-03-11
fix: preference dialog crash, missing ref to ACTION_DROP_IDX

nvandamme:


Shihabus-Sakib-Rad:
I had the same problem after installing it from AUR in cachyos. Changing `ACTION_DENY_IDX` to `ACTION_DROP_IDX` in `/usr/lib/python3.14/site-packages/opensnitch/dialogs/preferences/dialog.py` fixed it

PR #1606, opened 2026-05-21
daemon/rule: stop mutating regex Data field on Compile (#1587)

### LINKED ISSUE #1587: [Bug Report] Auto-generated regexps for appimage /tmp paths duplicate the a-z character class
### To Reproduce:

1. Download an AppImage, e.g. [Handy](https://github.com/cjpais/handy)
2. Launch it so it makes a request
3. Allow the request
4. Check the "From this executable"  path that was created

       ^/tmp/\.mount_handy_[0-9a-za-z]+\/.*handy$

5. Check the *name* of the rule - it will have the same duplication.

### Expected behavior (optional):

The regexp should presumably include `[0-9a-zA-Z]`, but OpenSnitch may pass the operand lowercase (even for paths, which may not be correct on case-sensitive filesystems, but that'd be a separate issue)

### Screenshots:

Image

### See also

https://github.com/evilsocket/opensnitch/issues/1392#issuecomment-3090996081

This was so easy to miss that I ended up with tens of rules containing `a-za-z`:

Image

tredondo:
Fixes #1587.

Compile() for Regexp operators with Sensitive==false used to do

    o.Data = strings.ToLower(o.Data)

before compiling the pattern. That works as an ASCII-only way to make matching case-insensitive (pattern and subject both lowercased), but it mangles the pattern string itself. A regex like [0-9A-Za-z]+ becomes [0-9a-za-z]+: still valid, but visibly wrong, and the mangled value then bleeds into:

  - the rule JSON written to disk by loader.Save (json.MarshalIndent serializes the post-Compile o.Data)
  - the GUI display of the operator data and the slugified rule name

This is most visible on the auto-generated rules for AppImages, whose process.path lives under /tmp/.mount_<name>_<random>/, producing the [0-9a-za-z]+ duplication the user can see in the popup and on disk.

Switch to Go regexp's native case-insensitivity instead: prepend the (?i) flag to a local pattern string and compile that. o.Data is no longer touched by Compile, so the original pattern survives both serialization and the round-trip back to the UI. reCmp can drop its matching strings.ToLower(data) call, since the compiled regex now handles case folding itself.

Existing rules on disk that already have a lowercased pattern keep working: (?i) over a [0-9a-z]+-style class is equivalent to the previous lowercase-both-sides behavior, and case-sensitive rules are unaffected (no (?i) is added).

A regression test (TestNewOperatorRegexpDataNotMutated) pins both the pattern-preservation contract and case-insensitive matching against an appimage-style path.

### DIFF
--- a/daemon/rule/operator.go
+++ b/daemon/rule/operator.go
@@ -172,10 +172,16 @@ func (o *Operator) Compile() error {
 		o.cb = o.rangeCmp
 	} else if o.Type == Regexp {
 		o.cb = o.reCmp
+		// Mark the compiled regex as case-insensitive via Go's (?i) flag
+		// instead of lowercasing o.Data. Lowercasing the pattern itself
+		// mangles character classes like [0-9A-Za-z] into [0-9a-za-z],
+		// which then leaks into the rule's saved JSON and the GUI.
+		// See https://github.com/evilsocket/opensnitch/issues/1587.
+		pattern := o.Data
 		if o.Sensitive == false {
-			o.Data = strings.ToLower(o.Data)
+			pattern = "(?i)" + pattern
 		}
-		re, err := regexp.Compile(o.Data)
+		re, err := regexp.Compile(pattern)
 		if err != nil {
 			return err
 		}
@@ -289,9 +295,8 @@ func (o *Operator) simpleCmp(v string) bool {
 }
 
 func (o *Operator) reCmp(data string) bool {
-	if o.Sensitive == false {
-		data = strings.ToLower(data)
-	}
+	// Case folding for the insensitive case is baked into the compiled regex
+	// via the (?i) flag added in Compile(), so the subject is matched as-is.
 	return o.re.MatchString(data)
 }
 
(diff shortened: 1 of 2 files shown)


munzzyy:
Built this against the pinned CI toolchain (Go 1.23, `protoc-gen-go@v1.34.1`, `protoc-gen-go-grpc@v1.3.0`) and ran the full `daemon/rule` package: `go build ./rule/...` is clean, and `go test ./rule/...` passes end to end (136s), including `TestNewOperatorRegexpDataNotMutated` and the existing `TestNewOperatorRegexpSensitive`. `(?i)` on a local pattern string instead of lowercasing `o.Data` in place is the right call - it keeps `Operator.Data` byte-for-byte what the user/UI sent, so the saved JSON and anything reading it back stays correct.

One thing worth flagging so it doesn't get lost: this fixes the regexp saved on disk/round-tripped to the GUI, but the UI's auto-generated rule *name* is a separate code path (`get_rule_name()` in `ui/opensnitch/dialogs/prompt/utils.py`) that slugifies the pattern client-side before the rule ever reaches the daemon. Since `slugify()` always lowercases, that name-generation step will still collapse `[0-9A-Za-z]` into `0-9a-za-z` even with this fix merged - it's the part of the original report showing "tens of rules containing a-za-z" in the rules list. Opened #1634 for that piece; no overlap with this one, just flagging since both reference #1587.

PR #1608, opened 2026-05-27
ui: resolve usernames with getent for UID rules

tttjhgan:
## Summary
  - Use `getent passwd` to populate local UID rule users instead of `pwd.getpwall()`
  - Resolve typed UID usernames through `pwd.getpwnam`, local `getent`, and remote non-interactive SSH `getent`
  - Prevent `uidCombo` from handling invalid indexes when users type custom text
  - Fix local node detection to compare IP addresses exactly instead of substring matching

  ## Testing
  - `python3 -m py_compile ui/opensnitch/dialogs/ruleseditor/dialog.py ui/opensnitch/nodes.py ui/tests/dialogs/test_ruleseditor.py ui/tests/
  test_nodes.py`
  - `git diff --check`

  Note: targeted pytest collection could not run in this local environment because the Python `grpc` module is missing.

### DIFF
--- a/ui/opensnitch/dialogs/ruleseditor/dialog.py
+++ b/ui/opensnitch/dialogs/ruleseditor/dialog.py
@@ -5,6 +5,7 @@ from datetime import datetime
 import sys
 import os
 import pwd
+import subprocess
 import time
 import ipaddress
 
@@ -49,7 +50,7 @@ class RulesEditorDialog(QtWidgets.QDialog, uic.loadUiType(DIALOG_UI_PATH)[0]):
         self._rules = Rules.instance()
         self._notification_callback.connect(self.cb_notification_callback)
         self._old_rule_name = None
-        self._users_list = pwd.getpwall()
+        self._users_list = []
 
         self.setupUi(self)
         self.setModal(modal)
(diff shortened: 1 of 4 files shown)


gustavo-iniguez-goya:
hi @tttjhgan ,

What was the motivation to change the current way of obtaining the users? any issues?
For me, running an external binary is a no-no by default, specially when `getent passwd` just returns the content of /etc/passwd.

tttjhgan:
> Hi @gustavo-iniguez-goya ,
>
> Thank you for the feedback. I’d like to clarify the use case behind this change.
>
> My working environment uses **FreeIPA**, with users synchronized from **Active Directory**. These are domain/centralized users managed through NSS/SSSD rather than static local accounts in `/etc/passwd`.
>
> The current implementation using `pwd.getpwall()` only returns local users from `/etc/passwd`. As a result, the synchronized FreeIPA/AD users cannot be listed or resolved in the UI, even though they have valid UIDs and can run processes on the system.
>
> This prevents administrators from creating UID-based network rules for those domain users, which is an important use case in enterprise environments.
>
> `getent passwd` follows the system NSS configuration (`/etc/nsswitch.conf`), so it can return both local users and users provided by sources such as SSSD, FreeIPA, LDAP, or AD integration. This allows OpenSnitch to resolve the actual UID/user mappings visible to the operating system.
>
> I understand the concern about invoking an external binary. The intention is not to replace the local-user behavior, but to use the standard NSS-aware mechanism so that UID rules also work correctly for centralized identity environments. The command can be invoked without shell interpolation, and user-provided values can be passed as arguments directly to avoid command-injection risks.
>
> In short, this change is needed because reading `/etc/passwd` alone is not sufficient for systems using centralized identity management.

PR #1616, opened 2026-06-08
fix "font rendering" due to 000-allow-localhost6.json 's description

workturnedplay:
the description for this rule, in the UI, vertically centers the 3 lines(2 text and one empty newline in the middle) thus showing only a few pixels of the first and last lines' text as the description for this rule making you believe there's something wrong with the font rendering.

before:
before

after:
 Removing one `\n` allows the two text lines to remain visible enough to be readable.
after


gustavo-iniguez-goya:
interesting... thank you!
What we'll need to investigate is how to limit rows content to 2 lines of text, in order not to cause these font rendering issues.

Issue #1624, opened 2026-06-17
[Feature Request] GUI add keyboard friendliness for tabs


kudashevs:
I would like to propose to improve GUI keyboard friendliness and add shortcuts for tabs. 

### Summary:
There should be the possibility to navigate between tabs with "Ctrl+Tab" and "Ctrl+Shift+Tab".

PR #1625, opened 2026-06-17
ui: add navigation between tabs

kudashevs:
In continuation to the proposal #1624 this is a PR with a possible implementation of the functionality.

This PR adds the possibility to navigate between GUI tabs with "Ctrl+Tab" and "Ctrl+Shift+Tab".

**Note** The only issue right now is when the GUI loses focus from tabs (i.e. focuses on a table in a tab) the shortcuts start navigating through the data of the focused table, not the tabs. If the functionality looks appropriate and will be accepted, I will find the solution for this issue (there is the possibility to get the focus back with esc or something like this.)


gustavo-iniguez-goya:
thank you! super useful

PR #1650, opened 2026-08-15
cli: add opensnitch-cli, a headless client for machines with no graphical environment

suinswofi:
## Summary

`opensnitch-cli` is a second implementation of the `UIServicer` for machines with no graphical environment. On a server the daemon only intercepts while a client is connected, and with none attached it falls back to `clientDisconnectedRule.Action` — so an unattended server allows everything. This fills that gap. It is an addition: nothing about `opensnitch-ui` changes for people who use it, and the two share their rule-building code rather than duplicating it.

- **`opensnitch-cli serve`** runs as a service and answers the daemon. The daemon holds the packet in a netfilter queue while it waits, so `serve` cannot wait for a person: it answers `AskRule` immediately with a short temporary rule (default `deny` for `1h`, fail-closed; `policy.unreviewed_action = allow` inverts it) and records the connection in a SQLite review queue.
- **`opensnitch-cli review`** goes through that queue afterwards, one connection at a time, and turns entries into real rules. It offers the same choices as the pop-up (executable, command line, host, domain wildcard, address, prefix, network, port, user, checksum, AppImage/snap paths, "also require" conditions), because both are built from the same code.
- **`rules`, `rule delete|enable|disable`, `undo`, `pending`, `nodes`, `status`** for managing what's there afterwards. `status` exits non-zero when the daemon rejected a rule, so it works as a monitoring check.

`serve` and `review` are separate processes that share only the database: `review` writes decisions to an outbox, `serve` sends them to the daemon within a second — so a decision taken while the service is stopped, or while the daemon is down, is applied when both are back.

Full documentation: `ui/opensnitch/cli/README.md`.

## Design notes for reviewers

- **Nothing under `opensnitch/cli/` imports PyQt.** `ui/tests/cli/test_no_qt.py` imports the whole package in a fresh interpreter and fails if anything reaches Qt. Making that possible is what the first four `ui:` commits do: `opensnitch/proto/__init__.py` reads the protobuf version directly instead of via `opensnitch.utils`; rule constants move to `opensnitch/rule_consts.py` (`Config` now inherits from it, so `Config.OPERAND_*` etc. keep working); the pop-up's operand builders move to `opensnitch/operands.py` and `dialogs/prompt/utils.py:get_combo_operator()` becomes a thin adapter over it. `ui/tests/cli/test_operands_parity.py` asserts the GUI and CLI produce identical `(type, operand, data)` for the same connection. That refactor also removed a real `config -> database -> utils -> config` import cycle that previously only resolved because some module happened to import `opensnitch.proto` first.
- **Why the temporary rule is withdrawn on approval.** `FindFirstMatch` lets any matching deny beat an allow, so a leftover temporary `cli-auto-*` deny would silently override a freshly approved allow. Approving a durable rule therefore also closes every queued entry it covers and withdraws their temporary rules (`opensnitch/cli/match.py` conservatively mirrors `operator.go` — it answers "can't tell" for anything it can't fully evaluate, never guesses), and `serve` answers a daemon asking about a covered connection with the not-yet-delivered decision itself instead of a new provisional rule.
- **`rule enable/disable` sends the whole rule back**, not just its name. `handleActionEnableRule` deserializes what it receives and `Replace()`s the stored rule with it, and `NewOperator` never fails, so a name-only rule leaves the daemon holding an empty one. (The GUI's events view currently sends name-only ENABLE/DISABLE — worth a look separately.)
- **`once` is refused** on the notifications channel: `addUserRule` skips it only on the Ask path; over `CHANGE_RULE` it is stored and never scheduled for removal.
- **Version guard.** The `Action` enum was renumbered in 1.6.0; a 1.5.x daemon does not reject this client's notifications, it *misreads* them (`CHANGE_RULE` → its `MONITOR_PROCESS`). `serve` warns loudly on Subscribe when the daemon is older than 1.6.0, and `nodes` shows each daemon's version. Ubuntu 24.04 still ships 1.5.8.
- **Packaging** is deliberately untouched beyond adding the script to `setup.py` `scripts`: `setup.py` declares no dependencies, so `pip install` pulls in no Qt. The service unit and example config live under `ui/resources/` for packagers rather than in `data_files`.

## Test plan

- `cd ui/tests && pytest -v cli/` — **215 passed, 1 skipped**, no Qt and no display needed. Includes an end-to-end test that runs the real gRPC server on a real unix socket with a stub daemon doing what `opensnitchd` does (subscribe, notifications stream, ping, ask, reply).
- Exercised against a daemon built from this tree (1.9.0) on Linux Mint 22.3: connect, deny/queue, review-approve, rule reaches `/etc/opensnitchd/rules`, survives daemon restart, decision taken while `serve` was stopped is delivered on start, `rule disable/delete` and `undo` round-trip, TCP-vs-unix node identity across a daemon reconnect.
- The GUI test suite (`ui/tests`, needs Qt) was not run in this environment; the `ui:` refactor commits were checked statically — every `Config.*` and `constants.*` name still referenced by the GUI resolves.

### DIFF
--- a/ui/bin/opensnitch-cli
+++ b/ui/bin/opensnitch-cli
@@ -1,0 +1,25 @@
+#!/usr/bin/env python3
+#
+#   Copyright (C) 2026 The OpenSnitch Authors
+#
+#   This file is part of OpenSnitch.
+#
+#   OpenSnitch is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   OpenSnitch is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with OpenSnitch.  If not, see <http://www.gnu.org/licenses/>.
+
+import sys
+
+from opensnitch.cli.main import main
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/ui/opensnitch/cli/README.md
+++ b/ui/opensnitch/cli/README.md
@@ -283,10 +283,12 @@ because the daemon lowercases it and it would never match.
 | `drop ID` | remove from the queue without creating a rule |
 | `rules [--node N] [--json]` | rules the daemon reported when it connected |
 | `nodes [--json]` | daemons that have connected |
-| `status [--json]` | queue depth, nodes, and rules the daemon rejected |
+| `status [--json] [--retry\|--clear]` | queue depth, nodes, and rules the daemon rejected |
 
 `status` exits non-zero if any rule was rejected, so it works as a monitoring
-check.
+check. `--retry` sends the rejected rules once more; `--clear` forgets them and
+puts their connections back in the review queue, so the decision can be taken
+again differently.
 
 `--match` takes an operand name, for example:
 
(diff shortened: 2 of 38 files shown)


suinswofi:
@gustavo-iniguez-goya @evilsocket — opening this as a draft rather than a formal PR on purpose. It's a fairly large addition (a headless client for servers, ~6.5k lines including tests and docs), so before asking for a review I'd like to know whether it's something you'd want in the tree at all, and if so whether the shape is right: a second `UIServicer` under `ui/opensnitch/cli/` sharing the pop-up's rule-building code, with nothing about `opensnitch-ui` changing for people who use it.

If it's welcome, I'll turn it into a proper PR and split or squash the commits however you prefer. If it's not something you want to maintain, no worries.

The description above has the design notes; `ui/opensnitch/cli/README.md` in the branch is the full documentation.

